PR #9, opened 2026-01-30
fix: use transport peer ID as P2PEvent source for correct message routing

mickvandijke:
## Summary
- P2PEvent::Message was using the logical `from` field from the JSON message payload as the source, but `send_message()` looks up peers by their transport-level peer ID
- This mismatch meant consumers receiving a P2PEvent::Message could not reply using the source field
- Now converts the transport `PeerId` to the hex string used in the peers map and uses that as the event source

## Test plan
- [x] Verify `cargo test` passes
- [x] Verify `cargo clippy -- -D warnings -D clippy::unwrap_used -D clippy::expect_used` passes
- [x] Confirm round-trip message routing works with the corrected peer ID

### DIFF
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -64,6 +64,7 @@ async fn main() -> Result<()> {
 
     // Create and start the node
     let node = P2PNode::new(config).await?;
+    node.start().await?;
 
     // Handle bootstrap peers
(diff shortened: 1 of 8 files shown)


mickvandijke:
@claude please review this PR

PR #18, opened 2026-02-06
feat: actually working closest peers

grumbach:
Implement working closest-peer discovery + fix peer ID comparison in DHT test

```js
Before (local + one-hop best effort)

  You
   |
   |-- local routing table
   |-- connected peers
   |-- query a few known peers (one hop)
   |
   +--> closest set they knew (most likely NOT actual closest on a large network)

After (iterative network lookup)

  You
   |
   |-- local routing table
   |-- connected peers
   |
   +--> query closest peers
         |
         +--> they return closer peers
                |
                +--> query those peers
                       |
                       +--> repeat until convergence
                       ...
                       +--> get ACTUAL closest nodes
```

  Summary

  - Implement a functional closest-peers workflow in the DHT, including manager/network/transport
    updates and comprehensive test coverage.
  - Add transport_peer_id() to DhtNetworkManager and update peer address population test to compare
    the correct peer ID format.

  What changed

  - Core DHT logic now returns working closest peers and supports iterative lookup behavior. (used to be one hop only!!!!!)
  - Transport/network integration updated to support the new closest-peers flow.
  - Added new DHT discovery/lookup tests and expanded peer address population coverage.
  - Fixed peer ID comparison in dht_peer_address_population_test by using transport IDs instead of
    config names.

  Tests added/updated

  - tests/dht_cross_node_discovery_test.rs
  - tests/dht_peer_address_population_test.rs
  - tests/iterative_dht_lookup_test.rs

  If you want a more detailed PR template (motivation, design notes, risks, rollout), tell me your
  preferred format.

## Follow-up: Stop conditions, trust hooks & geo-safe parallelism

- Candidate queues are now true FIFO windows with duplicate suppression. When the queue hits `MAX_CANDIDATE_NODES` we drop the newest entries instead of replacing older K-bucket members, which keeps the long-lived peers that passed geo/trust validation in place.
- Iterations snapshot the queued peer IDs; if the snapshot repeats we abort the lookup so we will not spin forever when remote peers keep returning the exact same candidates.
- Every response (value hit, closer-nodes reply, timeout) now feeds back into `EigenTrust` via the node's `report_peer_success/report_peer_failure` helpers. This keeps the trust layer informed about iterative lookups without adding any panics.
- Address parsing no longer assumes IPv4. Socket strings (with or without four-word suffixes) are normalized and rendered as `/ip4` or `/ip6` multiaddrs before they enter the phonebook, preserving geographic routing.
- Added doc notes in ADR-005 to capture the new safeguards and call out that all ALPHA-parallel queries share the same ant-quic socket (connection multiplexing), so we still comply with the transport-layer geo policies.

## Verification

- `cargo fmt`
- `cargo clippy --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used`
- `cargo test --all-features`

### DIFF
--- a/src/dht/core_engine.rs
+++ b/src/dht/core_engine.rs
@@ -469,11 +469,11 @@ const MAX_PARALLEL_QUERIES: usize = 3;
 /// K parameter - replication factor
 const K: usize = 8;
 
-/// Maximum value size for DHT store operations (1 MB)
+/// Maximum value size for DHT store operations (512 bytes)
 /// The DHT is designed as a "phonebook" for peer discovery, not general storage.
-/// Record types (NODE_AD, GROUP_BEACON, DATA_POINTER) should fit within this bound.
+/// Record types (NODE_AD, GROUP_BEACON, DATA_POINTER) should fit within 512 bytes.
 /// Larger data should be stored via send_message() in the application layer.
-const MAX_DHT_VALUE_SIZE: usize = 1_048_576;
+const MAX_DHT_VALUE_SIZE: usize = 512;
 
 /// Maximum node count for FindNode requests
 /// Prevents amplification attacks by limiting response size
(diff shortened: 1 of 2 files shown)


grumbach:
@claude please review

PR #19, opened 2026-02-06
fix: unify key derivation, sorted shortlist, and convergence detection

dirvine:
## Summary
- **Unified SHA-256 key derivation**: Replaced 7 zero-padded UTF-8 key derivation sites with canonical `peer_id_to_dht_key()` / `peer_id_to_key_bytes()` functions, ensuring consistent XOR distance calculations across the entire codebase
- **Sorted shortlist**: Replaced FIFO `VecDeque` candidate queue with `BTreeMap<distance, DHTNode>` via shared `IterativeLookup` struct, ensuring closest unqueried nodes are always selected first (standard Kademlia algorithm)
- **Convergence detection for `get()`**: Added distance-based convergence check that was previously only in `find_closest_nodes_network()` — lookups now terminate when no closer nodes are discovered
- **GetNotFound trust fix**: No longer records `record_peer_success` for "key not found" responses (neutral event, not a positive trust signal)
- **Module-level constants**: Promoted `ALPHA` (3) and `MAX_ITERATIONS` (20) from local bindings to module-level constants
- **Lookup dial timeout**: Added `LOOKUP_DIAL_TIMEOUT` (5s) to prevent unreachable nodes from blocking lookups
- **Reduced log noise**: Downgraded hot-path logging from `info!` to `debug!`/`trace!` in iterative lookup paths, request handlers, and message processing
- **Visibility fix**: Changed `DHTNode.cached_dht_key` from `pub` to `pub(crate)`
- **Test fixes**: Updated tests using values exceeding 512-byte `MAX_VALUE_SIZE` limit
- **ADR-005 update**: Reflects sorted shortlist algorithm and unified key derivation

## Test plan
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy -- -D warnings` passes (zero warnings)
- [x] `cargo doc --all-features --no-deps` passes
- [x] `cargo nextest run` — 1068/1069 pass (1 flaky timing-dependent test passes when run alone)
- [x] All iterative lookup tests pass (15/15)
- [x] All cross-node discovery tests pass (10/10)
- [x] All cross-node tests pass (7/7)

### DIFF
--- a/docs/adr/ADR-005-skademlia-witness-protocol.md
+++ b/docs/adr/ADR-005-skademlia-witness-protocol.md
@@ -272,16 +272,19 @@ pub enum ConsistencyLevel {
 
 To keep iterative lookups aligned with the multi-layer architecture, the DHT network manager now enforces:
 
-- **FIFO candidate queues**: new nodes are appended to a bounded queue (Kademlia-style K-buckets) and duplicates are ignored. When the queue hits `MAX_CANDIDATE_NODES` we drop the newest entrants, preserving the oldest, better-observed peers.
-- **Stagnation detection**: each iteration snapshots the candidate set; if the next iteration would query the identical peer set, the lookup terminates early instead of looping forever.
-- **Trust feedback hooks**: every successful response (value or closer nodes) reports a positive event to EigenTrust, while failures/timeouts register negative events. This keeps the trust layer informed without leaking panic paths.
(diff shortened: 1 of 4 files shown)


dirvine:
> src/dht_network_manager.rs:436 - BTreeMap collision bug: nodes with identical XOR distance will overwrite each other, potentially losing valid candidates during lookup

This from @greptile is not a bug, in an XOR network distance between nodes is unique 

@greptile @claude Please check my thinking here regarding the bug report

PR #87, opened 2026-04-15
feat(network): skip post-bootstrap self-lookups in client mode

mickvandijke:
## Summary
- Client-mode nodes (`NodeMode::Client`) now skip the two consecutive post-bootstrap self-lookups in `P2PNode::start()`. Full nodes (`NodeMode::Node`) keep the existing behaviour.
- Client startup previously spent ~100+ seconds in `dht_manager().trigger_self_lookup()` calls whose purpose is to refresh the *close* neighbourhood for peers that serve DHT queries. A client that only routes its own lookups doesn't need that — `bootstrap_from_peers` already seeds enough coverage to resolve its requests.

## Motivation
Measured cold-start of an `ant file download` (4 MB file, 3 chunks) against the public network:

| | Before | After this PR |
|---|---|---|
| Bootstrap dial | 53 s | 50 s |
| DHT peer discovery | (folded into self-lookup) | 20 s |
| Post-bootstrap self-lookups | **117 s** | **0 s** (skipped) |
| Download | 15 s | 102 s *(tall pole moved, see separate PR on lookup latency)* |
| Total wall | ~186 s | ~173 s |

The self-lookup cost was dominated by a handful of α-triples of FindNode queries all landing on NAT'd or dead peers whose dial cascades took ~30 s each. Skipping the two lookups entirely eliminates that phase for clients.

## Test plan
- [x] `cargo check --all-targets` on `saorsa-core` and the downstream `ant-core` / `ant-cli` workspace
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] End-to-end `ant file download` against public network: `Skipping post-bootstrap self-lookups (client mode)` logged at DEBUG, file retrieved successfully.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This PR gates the two post-bootstrap self-lookup rounds on `NodeMode::Node`, skipping them entirely for `NodeMode::Client` nodes and logging a debug message instead. The change is minimal, well-commented, and correctly preserves all surrounding bootstrap logic (peer connection, `bootstrap_from_peers`, bootstrapped flag, and close-group cache) for both modes.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is a targeted, low-risk optimization with no correctness or safety concerns.

[...]

PR #88, opened 2026-04-15
perf(dht): cut Kademlia lookup latency on cold clients

mickvandijke:
## Summary
Two related changes in `dht_network_manager.rs` that together cut a measured cold-start `ant file download` from ~173 s to ~97 s (-44%):

1. **`perf(dht)`: bound iteration wait after first response.** Each lookup iteration fires α concurrent queries and was waiting for all of them (`join_all`). Because `send_dht_request` does the full dial cascade inline, one NAT'd candidate could stall the iteration for 20–30 s — far past the point where Kademlia already has enough new candidates to advance. Drive the α queries through `FuturesUnordered`: block for the first response, then bound stragglers with `ITERATION_GRACE_TIMEOUT_SECS`. Any still-pending futures are dropped when the stream goes out of scope.

2. **`perf(dht)`: tune the grace to 5 s.** Initial value of 10 s turned out to fire on most iterations of a slow lookup — five consecutive iterations each took exactly ~10 s waiting on dead peers. Dropping to 5 s halves that worst case while still leaving ample slack for a straggler with an already-open channel.

## Motivation
Measured on repeated `ant file download` runs of the same 4 MB, 3-chunk file against the public network:

| Phase | Baseline | With grace 10 s | With grace 5 s |
|---|---|---|---|
| Bootstrap connect | 50 s | 51 s | 56 s |
| DHT peer discovery | 20 s | 16 s | 16 s |
| Self-lookups* | 0 s | 0 s | 0 s |
| Download (3 chunks) | 102 s | 73 s | **26 s** |
| **Total wall** | 173 s | 140 s | **~97 s** |

*Self-lookup skip is a separate PR (#87).

Per-iteration evidence (slow chunk, grace 10 s → 5 s): five consecutive iterations that capped at 10.2–11.2 s each dropped to near-instant iterations, cutting the slow chunk from 73 s to ~26 s.

## Test plan
- [x] `cargo check --all-targets` on `saorsa-core` and the downstream `ant-core` / `ant-cli` workspace
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] Four successive `ant file download` runs against public network, confirming the reductions in the table above.

### DIFF
--- a/src/dht_network_manager.rs
+++ b/src/dht_network_manager.rs
@@ -29,6 +29,7 @@ use crate::{
     network::NodeConfig,
 };
 use anyhow::Context as _;
+use futures::stream::{FuturesUnordered, StreamExt};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::net::{IpAddr, SocketAddr};
@@ -1232,10 +1233,22 @@ impl DhtNetworkManager {
                 batch.len()
             );
 
-            // Query nodes in parallel
-            // saorsa-transport connection multiplexing lets us keep a single transport socket
-            // while still querying multiple peers concurrently.
-            let query_futures: Vec<_> = batch
+            // Query nodes in parallel.
+            //
+            // saorsa-transport connection multiplexing lets us keep a single
+            // transport socket while still querying multiple peers
+            // concurrently.
+            //
+            // We drive the α queries through `FuturesUnordered` so we can
+            // advance the lookup as soon as there's *something* to work
+            // with. Waiting for every query (`join_all`) lets a single dead
+            // peer — whose dial cascade can take 20–30s — block the whole
+            // iteration; instead, once the first response arrives, we bound
+            // the wait on the stragglers to `ITERATION_GRACE_TIMEOUT_SECS`
+            // and move on with whatever responses came in by then. Any
+            // still-pending queries are dropped (and their futures cancelled)
+            // when the stream goes out of scope.
+            let query_stream: FuturesUnordered<_> = batch
                 .iter()
                 .map(|node| {
                     let peer_id = node.peer_id;
@@ -1259,7 +1272,7 @@ impl DhtNetworkManager {
                 })
                 .collect();
 
-            let results = futures::future::join_all(query_futures).await;
+            let results = Self::collect_iteration_results(query_stream).await;
 
             for (peer_id, result) in results {
                 queried_nodes.insert(peer_id);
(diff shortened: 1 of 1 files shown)


jacderida (approved):

grumbach (approved):
This is a really nice optimization. The core insight is correct: `join_all` on α queries means one dead peer (with a 20-30s dial cascade) blocks the entire iteration even though the other α-1 responses already give us enough candidates to proceed. Switching to `FuturesUnordered` + a 5s grace after first response is textbook async optimization for this pattern.

The implementation looks solid:
- `collect_iteration_results` cleanly separates the drain logic
- First response is always awaited (correct — no info = no progress)
- Grace timeout drops remaining futures which cancels in-flight dials
- 5s grace is generous enough for connected peers (<1s) while abandoning cascades that are almost certainly failing

One question: when the grace timeout fires and we drop the remaining futures, does the transport layer handle the mid-flight dial cancellation gracefully? If a QUIC handshake is in progress and the future is dropped, does saorsa-transport clean up the half-open connection? I'd expect it does (quinn drops connections on future cancellation), but worth confirming there's no leaked connection state.

This should noticeably improve the quote collection phase for uploads on NATted networks. Nice work.

mickvandijke:
@grumbach thanks for the careful read. Traced the cancellation path end-to-end — no leaked state.

**QUIC side is clean.** `dial_candidate` → `transport.connect_peer` → `connect_quic_async` → `NatTraversalEndpoint::connect_to` → quinn's `Connecting` future (saorsa-transport `nat_traversal_api.rs:3404`). Dropping `Connecting` is quinn's documented cancel path: it drops the underlying `Connection`, whose `Drop` impl sends CONNECTION_CLOSE (or the remote times the handshake out). saorsa-transport itself already wraps `connect_to` in `tokio::time::timeout`, so it's assuming exactly this cancel-safety.

**One small wrinkle in saorsa-core, already self-healing.** `send_dht_request` inserts a `DhtOperationContext` (holding the oneshot `Sender`) into `active_operations` *before* the dial (`dht_network_manager.rs:1820`). If we cancel mid-dial, the happy-path `ops.remove(&message_id)` never runs and the entry lingers. Defence-in-depth already lives at `dht_network_manager.rs:1740`: `sweep_expired_operations` runs at the top of every `send_dht_request` with a 2× timeout margin (~30s), so stale entries from dropped futures get reaped on the next lookup. Bounded, not a true leak.

**Hole-punch hints.** `dial_candidate` sets `hole_punch_target_peer_id` / `preferred_coordinator` on the transport keyed by socket addr before the dial (`dht_network_manager.rs:2055-2072`). Cancellation leaves those entries in place, but they're overwritten on the next dial to the same address — stale, not leaked.

Net: quinn handles the hard part and the existing sweep covers the bookkeeping. I'll add a line to the PR body noting this for future readers.

PR #89, opened 2026-04-16
perf(network): parallelize bootstrap peer dials with bounded concurrency

mickvandijke:
## Summary
- Phase A dials close-group cache peers serially to preserve trust-priority ordering (unchanged).
- Phase B dials CLI-configured and cached bootstrap peers concurrently via `futures::stream::buffer_unordered`, capped at `MAX_CONCURRENT_BOOTSTRAP_DIALS = 4` to bound simultaneous QUIC+PQC handshakes.
- Previously, a single dead peer at a 15s identity timeout could stall the entire bootstrap batch. With 4-wide concurrency, up to ~20 cached peers complete in ~5 waves worst case instead of sequentially.

## Test plan
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used -D clippy::expect_used` clean
- [x] `cargo test --lib bootstrap` — 18/18 pass
- [ ] Manual verification: cold-start a node with mixed live/dead bootstrap peers, confirm reduced time-to-first-connection

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This PR parallelizes Phase B of bootstrap peer dialing using `futures::stream::buffer_unordered` capped at `MAX_CONCURRENT_BOOTSTRAP_DIALS = 4`, reducing the worst-case cold-start latency from `N × 15 s` to `⌈N/4⌉ × 15 s` for dead peers. Phase A (close-group cache peers) remains serial to preserve trust-priority ordering.

The implementation is clean, well-commented, and correctly scoped: `dial_bootstrap_addr_set` takes `&self` so concurrent calls are safe, deduplication logic is unaffected, and all peer IDs are collected before running `bootstrap_from_peers`.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is a clean, well-bounded concurrency improvement with no correctness or safety issues.

All findings are P2 or lower. The parallelization correctly uses `buffer_unordered` with `&self` (no mutable aliasing), preserves Phase A trust ordering, and collects all peer IDs before DHT discovery. No panics, no unwraps, no security concerns introduced.

No files require special attention.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
[...]

PR #90, opened 2026-04-16
perf(network): prefer IPv4 over IPv6 when dialing peers

mickvandijke:
## Summary
Prefer IPv4 over IPv6 at every per-peer dial site. Stable sort preserves the caller's relative ordering within each family (trust-sort, XOR-distance, etc. stay meaningful).

Two changes:
- **`DhtNetworkManager::dialable_addresses`** sorts IPv4-first in one place, so every caller inherits the bias: dial loops (`dial_addresses`), and single-pick helpers (`first_dialable_address` and its downstream uses for hole-punch hints, routing-table touches, PublishAddress).
- **`P2PNode::connect_bootstrap_peers`** sorts each close-group and cached bootstrap peer's address set before the dial loop, since bootstrap builds its own address list rather than going through the DHT helper. CLI-configured peers are single-address sets so ordering is a no-op there.

`TransportHandle::connect_peer` receives one `MultiAddr` at a time from saorsa-core, so Happy Eyeballs racing inside the transport isn't on our call path.

## Rationale
IPv6 NAT traversal is less mature on consumer networks. When a peer advertises both families, an IPv4 attempt tends to succeed faster — and since dial loops break on the first successful identity handshake, preferring IPv4 reduces wasted timeouts during cold-start and DHT routing.

## Test plan
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings -D clippy::unwrap_used -D clippy::expect_used` clean
- [x] `cargo test --lib` — 282/282 pass
- [ ] Manual verification: cold-start a node against peers with mixed IPv4/IPv6 addresses, confirm IPv4 is attempted first per peer

## Notes
Independent of #89 (bootstrap dial parallelization). The sort lives inside `dialable_addresses` / bootstrap address-list construction, unaffected by #89's `serial_addr_sets` / `parallel_addr_sets` rename — either merge order rebases cleanly.

### DIFF
--- a/src/dht_network_manager.rs
+++ b/src/dht_network_manager.rs
@@ -1646,13 +1646,23 @@ impl DhtNetworkManager {
         queried.insert(self.config.peer_id);
     }
 
-    /// Return all dialable addresses from a list of [`MultiAddr`] values.
+    /// Return all dialable addresses from a list of [`MultiAddr`] values,
+    /// with IPv4 ordered before IPv6.
     ///
     /// Only QUIC addresses are considered dialable. Unspecified (`0.0.0.0`)
     /// addresses are rejected. Loopback addresses are accepted for local/test
     /// use.
+    ///
+    /// The IPv4-first preference centralises here so every caller — dial
+    /// loops (`dial_addresses`), single-pick helpers (`first_dialable_address`
+    /// and its downstream uses for hole-punch hints, routing-table touches,
+    /// PublishAddress, etc.) — inherits the same bias without duplicating the
+    /// sort. IPv6 NAT traversal is less mature on consumer networks, so IPv4
+    /// attempts tend to succeed faster when both families are available.
+    /// Stable sort preserves the caller's relative ordering within each
+    /// family (trust-sort, XOR-distance, etc. stay meaningful).
     fn dialable_addresses(addresses: &[MultiAddr]) -> Vec<MultiAddr> {
-        addresses
+        let mut filtered: Vec<MultiAddr> = addresses
             .iter()
             .filter(|addr| {
                 let Some(sa) = addr.dialable_socket_addr() else {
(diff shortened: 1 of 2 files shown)


grumbach:
Clean, well-scoped change. The one-place centralisation in `dialable_addresses` is the right move — `first_dialable_address` rides on top of it for free, so hole-punch hints / PublishAddress inherit the bias without a second sort to keep in sync.

**Blocking**

None.

**Polish**

- Address-family detection uses `MultiAddr::is_ipv4` which wraps `socket_addr().is_some_and(|a| a.is_ipv4())`. For the bootstrap path this is fine (we've already filtered to dialable QUIC), but in `dialable_addresses` the sort key runs on the already-filtered vec, so non-IP transports can't sneak in. Worth a one-line comment noting the sort assumes IP transports, otherwise a future reader adding a non-IP dialable transport will silently get it bucketed as "non-IPv4" which is accidentally correct but not obviously so.
- IPv4-mapped IPv6 (`::ffff:a.b.c.d`) will return `false` from `SocketAddr::is_ipv4` and land in the IPv6 bucket. Unlikely to appear in advertised peer sets (we construct `TransportAddr::Quic` from an `IpAddr`, and autonomi peers publish canonical forms), but if it ever does we'd dial it last. Non-issue in practice; flagging for completeness.
- `first_dialable_address` semantic shift for PublishAddress: we'll now advertise our IPv4 first. In an IPv6-only reachability scenario that's a minor regression, but given consumer-network reality it's a net win. Worth a line in the commit/PR body calling it out explicitly (the code comment covers it; the rationale doesn't).

**Nits**

- `sort_by_key(|a| !a.is_ipv4())` is correct and stable (Rust's `slice::sort_by_key` is guaranteed stable). No change needed — just confirming the doc comment's "stable sort preserves caller ordering" claim holds.
- Happy Eyeballs comment is accurate: `transport_handle::connect_peer` builds a single-element `addr_list` before calling `dual_node.connect_happy_eyeballs`, so no per-peer fan-out happens under us.

**Verdict**: Ship it. Together with ant-client #43 (IPv6 dual-stack default) this actually closes the IPv6-timeout loop — the client advertises both families, the dialer prefers the one that usually works. Low-risk, high-leverage.

grumbach (approved):

PR #110, opened 2026-05-04
transport: per-address reachability proofs + dial/send hardening

mickvandijke:
## Summary

Hardens the relay/reachability path and tightens transport timeout + send-failure semantics on top of saorsa-transport 0.34.0.

**Reachability — per-address Direct attribution**
- Replace global `direct_reachability_observed` with a per-pinned-external proof set, source-disjoint over observer IPs. An address is `Direct` only after `MIN_DISTINCT_OBSERVERS_FOR_DIRECT=2` peers (none in `dialed_addrs`, none in `known_peer_ips`) attribute an unsolicited inbound to it. Closes the pinhole-self-fulfillment case where a bootstrap redial flipped the global flag.
- Per-address attribution sourced from peers' QUIC `OBSERVED_ADDRESS` frames (new `P2pEvent::PeerObservedExternal`) instead of the per-family shortcut.
- Drop the live-observation fallback — `{direct,observed}_external_addresses()` now return only pinned, quorum-cleared addresses.

**Self-publish + dial plan shape**
- Publish Relay first, then one best non-Relay address per IP family (V4 + V6); same family priority Direct > Unverified > NATted, observed externals win same-tier ties.
- `select_dial_candidates` mirrors the publish shape: Relay + best per family, capped at three attempts.
- Retain unverified single-peer OBSERVED_ADDRESS reports as fallback publish candidates so relay-only self-records can advertise a NAT-mapped direct address.
- Promotion plumbing: `direct_address_promoted` mpsc from passive classifier → driver triggers an immediate typed-record republish on threshold crossing.
- Dedup typed self-address-set publishes; routine refreshes are skipped when neither the address set nor K-closest peers changed. State transitions (relay acquired/lost) bypass the cache.

**Transport — timeouts + send classification**
- Unify direct connect timeout into a single `direct_connect_timeout`; introduce `DIRECT_HANDSHAKE_TIMEOUT` (4s) separate from `DIRECT_CONNECT_TIMEOUT` (1s).
- New `SendFailureKind` distinguishes stale-channel (safe to reconnect/retry) from active-send (must not retry — would duplicate partial writes). `send_message` only retries on stale-channel.
- Drop the application-layer `connection_timeout` wrapper; transport's own open-stream / write-progress timeouts now drive timeouts.
- `connect_peer_typed` plumbs `AddressType` through dials so success/failure/timeout logs carry a structured `kind` field (Direct/Relay/Unverified/NATted).
- Normalize peer mappings; drop same-IP updates.

## Breaking changes

- `TransportHandle::direct_reachability_observed() -> bool` removed; replaced by `is_external_proven(SocketAddr) -> bool`.
- `SendFailureKind::PeerStopped` and `SendFailureKind::Acknowledgement` removed (saorsa-transport no longer emits them).
- Internal: `spawn_direct_reachability_classifier` and `spawn_peer_address_update_forwarder` take additional `Arc` parameters.

## Test plan

- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy --all-targets --all-features -- -D warnings`
- [ ] `cargo test --lib` (baseline 382 + 12 new = 394 expected)
- [ ] Manual: bootstrap redial against a node behind NAT does not flip `Direct` tag without two source-disjoint observers
- [ ] Manual: relay-only node publishes a NAT-mapped fallback once a peer reports it via `OBSERVED_ADDRESS`
- [ ] Manual: send during peer disconnect surfaces as active-send failure (no retry), stale-channel failure does retry

<h3>Greptile Summary</h3>

This PR replaces the global `direct_reachability_observed: AtomicBool` with a per-address proof system (`proven_externals: DashMap<SocketAddr, HashSet<IpAddr>>`), requiring at least 2 source-disjoint `OBSERVED_ADDRESS` reports to tag an external as `Direct`. It also introduces `SendFailureKind` to distinguish stale-channel (safe-to-retry) from active-send failures (must not retry), and overhauls the relay-driver publish shape to emit relay first then one best non-relay address per IP family.

- **P1** — `drain_direct_address_promoted()` and `drain_self_address_updated()` share a `tokio::sync::Mutex` with their `recv_*` counterparts; when the reachability driver is blocking inside a `select!` on `recv_direct_address_promoted()`, the `MutexGuard` is held across the await suspension, causing any concurrent `drain_*` caller to block indefinitely rather than performing the expected non-blocking `try_recv`.
- The per-address proof classifier, back-fill logic, and `PublishedTypedSet` dedup cache are well-designed and comprehensively unit-tested.

<h3>Confidence Score: 3/5</h3>

Not safe to merge as-is — the drain/recv mutex contention is a latent blocking bug that will surface once drain_* methods are used concurrently with the driver.

One P1 (drain_*/recv_* mutex starvation) and two P2s (register_connection_peer_id dual-stack regression, resolve_dial_address unconditional DHT lookup). The core per-address proof logic is well-reasoned and heavily tested, but the P1 is a subtle async correctness issue that warrants a fix before merging.

src/transport_handle.rs (drain_*/recv_* mutex), src/transport/saorsa_transport_adapter.rs (register_connection_peer_id early return), src/network.rs (resolve_dial_address DHT pre-fetch)

| Filename | Overview |
|----------|----------|
| src/transport_handle.rs | Core reachability overhaul: replaces global AtomicBool with per-address proven_externals DashMap, adds known_peer_ips/peer_observations/proof_eligible_peers state, and wires two new mpsc channels for promotion/update events. The drain_*/recv_* sharing of a single tokio::sync::Mutex is a design flaw that can cause indefinite blocking. |
| src/transport/saorsa_transport_adapter.rs | Large refactor: adds handle_peer_connected_for_proof, handle_peer_observed_external, back_fill_proof_on_pin, and classify_send_error helpers extracted into free functions for testability. Changes register_connection_peer_id to register only with the first endpoint that has an active connection (drops dual-stack alternate), and updates DIRECT_CONNECT_TIMEOUT from 2s to 1s with a new separate DIRECT_HANDSHAKE_TIMEOUT of 4s. |
| src/reachability/driver.rs | Driver now subscribes to recv_direct_address_promoted and recv_self_address_updated in both hold_until_lost and wait_backoff_or_event select loops; dedup publish cache (PublishedTypedSet) avoids redundant republishes; build_typed_self_address_set correctly emits relay first then one best non-relay address per IP family. |
| src/transport/external_addresses.rs | Adds unverified Vec and record_unverified() method alongside pinned-direct list; pin_direct now returns bool and promotes from unverified; normalization applied consistently in set_relay. Well-tested. |
| src/error.rs | New SendFailureKind enum distinguishes stale-channel from active-send failures; TransportError::SendFailed replaces the prior generic stream error; P2PError::is_stale_channel_send_failure drives retry policy. Well-designed and well-tested. |
| src/dht_network_manager.rs | select_dial_candidates rewritten to select one best non-relay address per IP family (capped at 3 total); dial_candidate now takes AddressType for structured logging; removes the bare peer_addresses_for_dial wrapper. |
| src/network.rs | resolve_dial_address now returns (MultiAddr, AddressType) and always pre-fetches DHT typed addresses upfront; reconnect fast-path correctly gates on is_stale_channel_send_failure. Unconditional DHT lookup is a minor performance regression. |
| src/reachability/session.rs | Adds per-candidate debug logging during relay acquisition; comment updates reflect new publish shape. No logic changes. |
| src/adaptive/dht.rs | Renames peer_addresses_for_dial to peer_addresses_for_dial_typed to surface the AddressType tag; thin delegation to the manager layer. |
| src/dht/core_engine.rs | Doc comment update only: references updated method name peer_addresses_for_dial_typed. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant PT as saorsa_transport
    participant AF as AddressForwarder
    participant RC as ReachabilityClassifier
    participant EA as ExternalAddresses
    participant PE as proven_externals
    participant AD as AcquisitionDriver

    PT->>RC: PeerConnected(addr, Side::Server)
    RC->>RC: source-disjoint / sibling-hairpin filters
    RC->>EA: direct_addresses() [for hairpin check]
    RC->>PE: credit_observation(peer_ip, ext) [if eligible]
    RC-->>AD: emit_direct_promotion(addr) [if threshold crossed]

    PT->>RC: PeerObservedExternal(peer_addr, observed_ext)
    RC->>EA: record_unverified(observed_ext)
    RC->>PE: credit_observation(peer_ip, observed_ext) [if eligible]
    RC-->>AD: emit_self_address_update(addr) [if new unverified]
    RC-->>AD: emit_direct_promotion(addr) [if threshold crossed]

    PT->>AF: ExternalAddressDiscovered(addr)
    AF->>EA: pin_direct(addr)
    AF->>PE: back_fill_proof_on_pin(addr, observations, eligible)
    AF-->>AD: emit_direct_promotion(addr) [if back-fill promotes]
    AF-->>AD: emit_self_address_update(addr) [if no promotion]

    AD->>AD: select recv_direct_address_promoted / recv_self_address_updated
    AD->>AD: publish_typed_set_with_policy(relay, force=false)
    AD->>AD: build_typed_self_address_set [Relay first, best per family]
    AD->>AD: dedup via PublishedTypedSet cache
```

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `src/transport_handle.rs`, line 1302-1313 ([link](https://github.com/saorsa-labs/saorsa-core/blob/9bcd4c85f181565d6982e5c6b5f7d0919b0484a7/src/transport_handle.rs#L1302-L1313)) 

   <a href="#">P1</a> **`drain_*` methods block indefinitely when the driver holds the shared mutex**

   `drain_direct_address_promoted()` and `drain_self_address_updated()` share the same `tokio::sync::Mutex<mpsc::Receiver>` as their `recv_*` counterparts. When the reachability driver has `recv_direct_address_promoted()` inside a `select!`, the `MutexGuard` is held across the `rx.recv().await` suspension point. Any concurrent call to `drain_direct_address_promoted()` will block on `.lock().await` indefinitely — until a promotion event actually arrives and the driver releases the guard — despite the method using `try_recv()` internally and appearing to be a quick non-blocking poll. This contradicts the API contract implied by the "drain" name and can stall callers unexpectedly.

   `````markdown
   Path: src/transport_handle.rs
   Line: 1302-1313

   Comment:
   **`drain_*` methods block indefinitely when the driver holds the shared mutex**

   `drain_direct_address_promoted()` and `drain_self_address_updated()` share the same `tokio::sync::Mutex<mpsc::Receiver>` as their `recv_*` counterparts. When the reachability driver has `recv_direct_address_promoted()` inside a `select!`, the `MutexGuard` is held across the `rx.recv().await` suspension point. Any concurrent call to `drain_direct_address_promoted()` will block on `.lock().await` indefinitely — until a promotion event actually arrives and the driver releases the guard — despite the method using `try_recv()` internally and appearing to be a quick non-blocking poll. This contradicts the API contract implied by the "drain" name and can stall callers unexpectedly.

   `````
   </details>

2. `src/transport/saorsa_transport_adapter.rs`, line 636-653 ([link](https://github.com/saorsa-labs/saorsa-core/blob/9bcd4c85f181565d6982e5c6b5f7d0919b0484a7/src/transport/saorsa_transport_adapter.rs#L636-L653)) 

   <a href="#">P2</a> **Early `return` after first endpoint drops dual-stack peer ID registration**

   The new logic registers the peer ID only with the first endpoint that reports `has_active_connection`, then immediately returns. Previously, the peer ID was registered with **both** the v6 and v4 endpoints (plus the IPv4-mapped alternate form), which ensured PUNCH_ME_NOW relay messages arriving on either stack could locate the peer.

   On a dual-stack node: if a peer connects via IPv4 (registered with v4 endpoint only), a relay coordination message arriving on the v6 endpoint will fail to find the peer ID mapping, breaking PUNCH_ME_NOW routing for that peer. The race window where `has_active_connection` returns false on both stacks (connection just established) would silently swallow the registration entirely.

   `````markdown
   Path: src/transport/saorsa_transport_adapter.rs
   Line: 636-653

   Comment:
   **Early `return` after first endpoint drops dual-stack peer ID registration**

   The new logic registers the peer ID only with the first endpoint that reports `has_active_connection`, then immediately returns. Previously, the peer ID was registered with **both** the v6 and v4 endpoints (plus the IPv4-mapped alternate form), which ensured PUNCH_ME_NOW relay messages arriving on either stack could locate the peer.

   On a dual-stack node: if a peer connects via IPv4 (registered with v4 endpoint only), a relay coordination message arriving on the v6 endpoint will fail to find the peer ID mapping, breaking PUNCH_ME_NOW routing for that peer. The race window where `has_active_connection` returns false on both stacks (connection just established) would silently swallow the registration entirely.

   `````
   </details>

3. `src/network.rs`, line 1622-1636 ([link](https://github.com/saorsa-labs/saorsa-core/blob/9bcd4c85f181565d6982e5c6b5f7d0919b0484a7/src/network.rs#L1622-L1636)) 

   <a href="#">P2</a> **Unconditional DHT lookup on every reconnect, even with caller-provided addresses**

   The old `resolve_dial_address` short-circuited immediately when caller-provided or saved addresses were available. The new version always fires `peer_addresses_for_dial_typed()` upfront to populate `dht_typed` for `lookup_kind`. For most reconnect flows (which have caller- or saved-addresses), this is an unnecessary async DHT round-trip on the critical reconnect path. Consider lazily fetching DHT results only when caller/saved addresses are exhausted, and defaulting the address type to `AddressType::Unverified` for non-DHT sources without querying the DHT first.

   `````markdown
   Path: src/network.rs
   Line: 1622-1636

   Comment:
   **Unconditional DHT lookup on every reconnect, even with caller-provided addresses**

   The old `resolve_dial_address` short-circuited immediately when caller-provided or saved addresses were available. The new version always fires `peer_addresses_for_dial_typed()` upfront to populate `dht_typed` for `lookup_kind`. For most reconnect flows (which have caller- or saved-addresses), this is an unnecessary async DHT round-trip on the critical reconnect path. Consider lazily fetching DHT results only when caller/saved addresses are exhausted, and defaulting the address type to `AddressType::Unverified` for non-DHT sources without querying the DHT first.

   `````
   </details>

   Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

</details>

`````markdown
Fix the following 3 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 3
src/transport_handle.rs:1302-1313
**`drain_*` methods block indefinitely when the driver holds the shared mutex**

`drain_direct_address_promoted()` and `drain_self_address_updated()` share the same `tokio::sync::Mutex<mpsc::Receiver>` as their `recv_*` counterparts. When the reachability driver has `recv_direct_address_promoted()` inside a `select!`, the `MutexGuard` is held across the `rx.recv().await` suspension point. Any concurrent call to `drain_direct_address_promoted()` will block on `.lock().await` indefinitely — until a promotion event actually arrives and the driver releases the guard — despite the method using `try_recv()` internally and appearing to be a quick non-blocking poll. This contradicts the API contract implied by the "drain" name and can stall callers unexpectedly.

### Issue 2 of 3
src/transport/saorsa_transport_adapter.rs:636-653
**Early `return` after first endpoint drops dual-stack peer ID registration**

The new logic registers the peer ID only with the first endpoint that reports `has_active_connection`, then immediately returns. Previously, the peer ID was registered with **both** the v6 and v4 endpoints (plus the IPv4-mapped alternate form), which ensured PUNCH_ME_NOW relay messages arriving on either stack could locate the peer.

On a dual-stack node: if a peer connects via IPv4 (registered with v4 endpoint only), a relay coordination message arriving on the v6 endpoint will fail to find the peer ID mapping, breaking PUNCH_ME_NOW routing for that peer. The race window where `has_active_connection` returns false on both stacks (connection just established) would silently swallow the registration entirely.

### Issue 3 of 3
src/network.rs:1622-1636
**Unconditional DHT lookup on every reconnect, even with caller-provided addresses**

The old `resolve_dial_address` short-circuited immediately when caller-provided or saved addresses were available. The new version always fires `peer_addresses_for_dial_typed()` upfront to populate `dht_typed` for `lookup_kind`. For most reconnect flows (which have caller- or saved-addresses), this is an unnecessary async DHT round-trip on the critical reconnect path. Consider lazily fetching DHT results only when caller/saved addresses are exhausted, and defaulting the address type to `AddressType::Unverified` for non-DHT sources without querying the DHT first.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(reachability): dedup typed self add..."](https://github.com/saorsa-labs/saorsa-core/commit/9bcd4c85f181565d6982e5c6b5f7d0919b0484a7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30695790)</sub>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main, develop ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'rc-*' ]
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main, develop ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'rc-*' ]
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main, develop ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'rc-*' ]
   workflow_dispatch:
 
 concurrency:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.34.0"
-source = "git+https://github.com/saorsa-labs/saorsa-transport.git?branch=rc-2026.4.4#4d0c4a868f208c288323c7b9a2c0d44c2f55173e"
+source = "git+https://github.com/saorsa-labs/saorsa-transport.git?branch=rc-2026.4.4#e0f7bcd2c7011f298a112f242656b2d3269aef43"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1088,9 +1088,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.13"
+version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
+checksum = "171fefbc92fe4a4de27e0698d6a5b392d6a0e333506bc49133760b3bcf948733"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -2637,9 +2637,8 @@ dependencies = [
 
 [[package]]
 name = "saorsa-transport"
-version = "0.33.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9df373519d88540825f435bf3c5c6e300377871b28f16cdd461db3c3f4bb5323"
+version = "0.34.0"
+source = "git+https://github.com/saorsa-labs/saorsa-transport.git?branch=rc-2026.4.4#4d0c4a868f208c288323c7b9a2c0d44c2f55173e"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3128,9 +3127,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.52.1"
+version = "1.52.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b67dee974fe86fd92cc45b7a95fdd2f99a36a6d7b0d431a231178d3d670bbcc6"
+checksum = "110a78583f19d5cdb2c5ccf321d1290344e71313c6c37d43520d386027d18386"
 dependencies = [
  "bytes",
  "libc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ once_cell = "1.21"
 dashmap = "6"
 
 # Networking
-saorsa-transport = "0.33.0"
+saorsa-transport = { git = "https://github.com/saorsa-labs/saorsa-transport.git", branch = "rc-2026.4.4" }
 
 # Core-specific dependencies
 dirs = "6.0"
--- a/src/adaptive/dht.rs
+++ b/src/adaptive/dht.rs
@@ -21,6 +21,7 @@
 //! trust signals through [`AdaptiveDHT::report_trust_event`].
 
 use crate::adaptive::trust::{NodeStatisticsUpdate, TrustEngine};
+use crate::dht::core_engine::AddressType;
 use crate::dht_network_manager::{DhtNetworkConfig, DhtNetworkManager};
 use crate::{MultiAddr, PeerId};
 
@@ -247,13 +248,20 @@ impl AdaptiveDHT {
         self.dht_manager.trigger_self_lookup().await
     }
 
-    /// Look up connectable addresses for a peer.
+    /// Look up connectable typed addresses for a peer.
     ///
-    /// Checks the DHT routing table first, then falls back to the transport
-    /// layer. Returns an empty vec when the peer is unknown or has no dialable
-    /// addresses.
-    pub(crate) async fn peer_addresses_for_dial(&self, peer_id: &PeerId) -> Vec<MultiAddr> {
-        self.dht_manager.peer_addresses_for_dial(peer_id).await
+    /// Checks the DHT routing table first, then falls back to the
+    /// transport layer. Returns an empty vec when the peer is unknown
+    /// or has no dialable addresses. The per-address [`AddressType`]
+    /// tag is preserved so the dial path can log the kind on
+    /// success/failure.
+    pub(crate) async fn peer_addresses_for_dial_typed(
+        &self,
+        peer_id: &PeerId,
+    ) -> Vec<(MultiAddr, AddressType)> {
+        self.dht_manager
+            .peer_addresses_for_dial_typed(peer_id)
+            .await
     }
 }
 
--- a/src/dht/core_engine.rs
+++ b/src/dht/core_engine.rs
@@ -1273,7 +1273,7 @@ impl DhtCoreEngine {
 
     /// Get a peer's addresses paired with their [`AddressType`] tags.
     ///
-    /// Used by [`crate::dht_network_manager::DhtNetworkManager::peer_addresses_for_dial`]
+    /// Used by [`crate::dht_network_manager::DhtNetworkManager::peer_addresses_for_dial_typed`]
     /// to sort candidates by type priority (Relay first) per ADR-014.
     pub async fn get_node_addresses_typed(
         &self,
--- a/src/dht_network_manager.rs
+++ b/src/dht_network_manager.rs
@@ -2141,7 +2141,7 @@ impl DhtNetworkManager {
         }
         let mut attempted = 0usize;
         let mut skipped_cached = 0usize;
-        for (addr, _ty) in &plan {
+        for (addr, ty) in &plan {
             attempted += 1;
             let Some(socket_addr) = addr.dialable_socket_addr() else {
                 continue;
@@ -2149,13 +2149,14 @@ impl DhtNetworkManager {
             if self.dial_failure_cache.is_failed(&socket_addr) {
                 skipped_cached += 1;
                 trace!(
-                    "dial_addresses: skipping recently failed address {} for {}",
+                    "dial_addresses: skipping recently failed address {} ({:?}) for {}",
                     addr,
+                    ty,
                     peer_id.to_hex()
                 );
                 continue;
             }
-            match self.dial_candidate(peer_id, addr).await {
+            match self.dial_candidate(peer_id, addr, *ty).await {
                 Some(channel_id) => {
                     self.dial_failure_cache.clear(&socket_addr);
                     return Some(channel_id);
@@ -2659,14 +2660,19 @@ impl DhtNetworkManager {
     /// [`TransportHandle::wait_for_peer_identity`] before sending, because
     /// the app-level `peer_to_channel` mapping is only populated after the
     /// asynchronous identity-exchange handshake completes.
-    async fn dial_candidate(&self, peer_id: &PeerId, address: &MultiAddr) -> Option<String> {
+    async fn dial_candidate(
+        &self,
+        peer_id: &PeerId,
+        address: &MultiAddr,
+        kind: AddressType,
+    ) -> Option<String> {
         let peer_hex = peer_id.to_hex();
 
         // Reject unspecified addresses before attempting the connection.
         if address.ip().is_some_and(|ip| ip.is_unspecified()) {
             debug!(
-                "dial_candidate: rejecting unspecified address for {}: {}",
-                peer_hex, address
+                "dial_candidate: rejecting unspecified address for {}: {} ({:?})",
+                peer_hex, address, kind
             );
             return None;
         }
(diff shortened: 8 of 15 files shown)


jacderida:
Addressed the review in `be3e7e9`. A note on each:

**P1 — `drain_*` / `recv_*` shared mutex (`src/transport_handle.rs`)**

Already fixed on the branch in `016f429` (one commit after the one Greptile reviewed at `9bcd4c85`). The `drain_direct_address_promoted` / `drain_self_address_updated` methods no longer share the `tokio::sync::Mutex<mpsc::Receiver<…>>` with their `recv_*` counterparts. A new `AddressEventPublisher` in `src/transport/saorsa_transport_adapter.rs` pairs the existing mpsc work-queue (single-consumer for the reachability driver) with a parallel `tokio::sync::watch::Sender`. The `drain_*` methods now read from the watch receiver via `drain_latest_socket_event` (`src/transport_handle.rs:1006`) under a synchronous `parking_lot::Mutex` — `has_changed()` + `borrow_and_update()`, no `.await` while the lock is held — so concurrent observers can never contend with the driver's `select!` suspension. Multiple raw events coalesce to the latest address, as the doc comments now state.

**P2 — `register_connection_peer_id` dual-stack regression (`src/transport/saorsa_transport_adapter.rs`)**

Greptile is right that the early `return` after the first matching endpoint dropped registration on the second stack and on the IPv4-mapped alternate form (the alternate was registered explicitly in `5e68590`, the commit that introduced peer-ID routing). I kept the `has_active_connection` gate that `68b9f73` introduced — it's correct that we should not register against a stack that doesn't own the connection — but removed the early return and re-added the alternate-form registration. Both stacks are now walked, and each is registered independently for the normalized form and (where active) for the dual-stack alternate. The race-window concern is largely moot in practice: the only caller (`src/transport_handle.rs:2316`) runs after parsing an authenticated message just received from the address, so the owning endpoint must have an active connection at that moment.

**P2 — `resolve_dial_address` unconditional routing-table read (`src/network.rs`)**

`peer_addresses_for_dial_typed` is a routing-table read (RwLock + DashMap), not a network round-trip, so the cost is small — but the result was only consumed as a log tag (`connect_peer_typed` plumbs it into the structured `kind` field; nothing behavioural). On the hot reconnect path, paying that cost just to label caller- / saved-supplied addresses isn't worth it. Caller- and saved-supplied addresses now default to `AddressType::Unverified` — the same fallback the routing table itself uses for legacy peers — and the DHT lookup only fires when both upstream sources are exhausted. The function's existing doc comment (`src/network.rs:1652-1661`) already described this `Unverified`-fallback behaviour, so the implementation now matches the contract.

**Verification note**

I couldn't run a full `cargo build` / clippy locally: the branch already has 11 pre-existing compile errors against `saorsa-transport 0.33.0` on crates.io (the code calls `0.34.0`-only APIs — `SendFailureStage`, `PeerObservedExternal`, `with_direct_connect_timeout`, `has_active_connection`, `EndpointError::SendFailed`). My edits introduce one additional `has_active_connection` use on the same root cause and will resolve in lockstep once the transport version is bumped. `cargo fmt --all -- --check` passes. The base branch's CI (`pull_request: branches: [main]`) doesn't trigger for this PR (which targets `rc-2026.4.4`), so the broken state isn't currently surfaced by checks.

jacderida:
Heads-up on a few extra commits that went on top of the review fix:

**Dep: switch saorsa-transport to a git reference**

The 0.34 transport surface this PR depends on (`PeerObservedExternal`, `SendFailureStage`, `with_direct_connect_timeout`, `has_active_connection`, `EndpointError::SendFailed`) isn't on crates.io yet — `0.33.0` is the latest published. Cargo.toml now points at the `rc-2026.4.4` branch over HTTPS so the build resolves. Worth swapping back to a published version once `0.34.0` lands.

**CI: trigger main checks on rc-* PRs**

`lint.yml`, `rust.yml`, `unit-tests.yml` were filtered to `pull_request: branches: [main]`, so PRs targeting `rc-2026.4.4` only got `claude-review` — actual lint, build, and unit-test gates were silent. Added `'rc-*'` alongside `main` in each workflow's `pull_request.branches` filter so those gates fire on RC PRs too.

(Same gap exists on saorsa-transport's CI — `[master, main]` only — and let the Windows breakage below sneak through. Worth mirroring the change there at some point but I left that as a follow-up.)

**Test: `wildcard_bind_publishes_observed_address_after_peer_connection`**

Failed on every platform. `local_dht_node()` (`src/dht_network_manager.rs:1995`) was reading `direct_external_addresses()` — Direct quorum-cleared only — so a node whose post-NAT address had been observed by just one peer published an empty self-entry. The relay-driver typed-publish path was already advertising the same address as Unverified (commit `23433e7`), but the synchronous self-entry never picked up the Unverified tier when the source-disjoint quorum landed in `0b17803` / `a147e4d`. Switched to `non_relay_external_addresses()` so the synchronous entry mirrors the typed-publish surface (Direct + Unverified, still no relay).

**Build: Windows libc::EMSGSIZE**

Surfaced by the new `'rc-*'` PR triggers above. `saorsa-transport`'s `is_message_too_large` (`src/masque/relay_server.rs`) referenced `libc::EMSGSIZE` unconditionally, but `libc` is declared as `[target.'cfg(unix)'.dependencies]`, so the Windows build hit `E0433: cannot find module or crate libc`. Fixed in `saorsa-labs/saorsa-transport@e0f7bcd2` (already merged to that repo's `rc-2026.4.4`): kept the libc-backed implementation behind `cfg(unix)`, added a `cfg(not(unix))` no-op fallback that returns `false` — same shape as the existing platform-gated `set_dont_fragment` helper a few lines up. Cargo.lock here pins the new transport commit.

All ten checks now pass:

| Check | Result |
|---|---|
| Build (ubuntu/macos/windows) | ✅ |
| Lint & Format | ✅ |
| Security Audit | ✅ |
| Unit Tests (Linux/macOS/Windows) | ✅ |
| claude-review | ✅ |

PR #115, opened 2026-05-09
Stabilize DHT lookups and transport sends

mickvandijke:
## Summary
- Hardens DHT lookup correctness by rejecting stale relay winners, propagating failed lookup peers, skipping peers with exhausted dial plans, handling lagged lookup-failure receivers, and avoiding liveness refresh from address gossip alone.
- Stabilizes self-address and transport behavior with structured relay/IPv4/IPv6 candidate selection, the current saorsa-transport stability branch, updated Cargo.lock pins, and adjusted QUIC direct-connect/send semantics.
- Enforces non-zero DHT address publish/replace sequence numbers so zero stays reserved as the no-sequence sentinel.
- Classifies LAN-scoped addresses separately from Direct, Unverified, and Relay addresses, replacing AddressType::NATted with AddressType::Lan while still accepting the legacy NATted wire name as Lan.
- Tightens LAN-first dialing by canonicalizing IPv4-mapped LAN addresses and requiring same-WAN compatibility before preferring private/link-local/loopback paths; otherwise peers fall back to WAN or relay candidates.

## Recent additions
- feat(network)!: classify LAN addresses separately
- fix(network): canonicalize mapped IPv4 LAN addresses
- fix(network): require same WAN for LAN-first dialing

## Testing
- Not run in this session (PR description only).

<h3>Greptile Summary</h3>

This PR hardens DHT lookup correctness and self-address stability across several independent concerns: stale relay records are no longer treated as authoritative results during iterative FIND_NODE lookups, peer failures observed by one active lookup are propagated to sibling lookups via a broadcast coordinator, and admission revalidation is moved off the transport event loop onto its own spawned task. A new `SelfAddressSet` struct replaces the earlier `Vec`+`HashSet` approach with structured relay/IPv4/IPv6 slots that produce a stable, deduplicated address set every time.

- `AddressReplaceMode` in `core_engine.rs` prevents third-party gossip from refreshing `last_seen` or bucket MRU order; `touch_legacy_relay_hint_if_unsequenced` blocks stale relay hints from overwriting a peer that has already published a sequenced address record.
- `compute_winner` gains a \"newest publish-sequence wins\" rule (before quorum and XOR-distance fallback), and `apply_lookup_report_winners` replaces stale candidate copies with the per-subject winner before returning results to callers.
- `saorsa-transport` is switched from the crates.io registry to a git branch for upstream stability fixes; `DeliveryAck` is mapped to `StreamFinish` (not `StaleChannel`) to avoid reconnect-triggered duplicate sends.

<h3>Confidence Score: 4/5</h3>

The core logic changes are sound and backed by a solid suite of new unit tests; the main open question is stabilising the upstream transport dependency before merging to a release branch.

All production code correctly avoids panics and the DHT correctness improvements are well-reasoned and tested. The broadcast-lag case in wait_for_lookup_failure_signal silently drops the early-abort signal, a minor performance concern rather than a correctness one. The saorsa-transport git branch reference is the more pressing concern: a force-push or deletion breaks any workflow that regenerates Cargo.lock.

Cargo.toml and Cargo.lock warrant a second look before landing on a release branch; src/dht_network_manager.rs is the largest change and benefits from end-to-end integration testing since the lookup failure propagation path has not been exercised in this session.

| Filename | Overview |
|----------|----------|
| Cargo.toml | saorsa-transport switched from pinned registry version (0.34.1) to a mutable git branch. Cargo.lock pins a specific commit today, but the branch reference is unstable and could cause build failures if force-pushed or deleted. |
| src/dht/core_engine.rs | Adds AddressReplaceMode enum to gate liveness refreshes, touch_legacy_relay_hint_if_unsequenced guard, replace_node_addresses_from_gossip gossip path, and find_/all_nodes_with_publish_seq query helpers. Logic is consistent and well-tested. |
| src/dht_network_manager.rs | Major rework: LookupPeerStates replaces HashSet, LookupFailureCoordinator broadcasts peer failures across concurrent lookups, compute_winner gains a publish-seq freshness rule, apply_lookup_report_winners refreshes stale candidate copies. Minor: Lagged broadcast receiver can silently miss the target failure signal, losing the early-abort optimisation without any indication. |
| src/network.rs | Relay hint path switched to touch_legacy_relay_hint_if_unsequenced, preventing stale relay hints from overwriting a sequenced publish record. Cosmetic variable rename for send_result. |
| src/reachability/driver.rs | Trivial call-site update: build_typed_self_address_set renamed to build_self_address_set returning SelfAddressSet; into_typed_vec() materialises the vec for the publish fan-out. No logic changes. |
| src/self_address.rs | SelfAddressSet replaces Vec+HashSet tracking with structured relay/v4/v6 slots that emit addresses in stable first-seen order. Dedup scope narrowed from all-seen to selected-only, which is correct because unselected same-family losers cannot displace the winner later. Well-tested. |
| src/transport/saorsa_transport_adapter.rs | DIRECT_CONNECT_TIMEOUT bumped from 1000ms to 1250ms to accommodate one QUIC initial-flight retransmission. Comment clarified for send() delivery semantics. Both changes are non-breaking. |
| src/transport_handle.rs | New DeliveryAck SendFailureStage variant mapped to StreamFinish (not StaleChannel), so delivery-ack failures do not trigger reconnect+retry — correct for QUIC semantics where the data may already have been received. Cosmetic variable rename only. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant DhtNetworkManager
    participant LookupPeerStates
    participant LookupFailureCoordinator
    participant RemotePeer

    Caller->>DhtNetworkManager: iterative_find_nodes(key)
    DhtNetworkManager->>LookupPeerStates: mark_self_queried (Succeeded)
    DhtNetworkManager->>LookupPeerStates: mark_failed (exhausted dial plans)
    loop Alpha batch
        DhtNetworkManager->>LookupPeerStates: mark_waiting(peer)
        DhtNetworkManager->>LookupFailureCoordinator: "subscribe() -> failure_rx"
        DhtNetworkManager->>RemotePeer: send_find_node_lookup_request
        alt Request completes with error
            RemotePeer-->>DhtNetworkManager: Err
            DhtNetworkManager->>LookupFailureCoordinator: notify_failed(peer)
            DhtNetworkManager->>LookupPeerStates: mark_failed(peer)
        else External failure signal arrives first
            LookupFailureCoordinator-->>DhtNetworkManager: peer failed (from sibling lookup)
            DhtNetworkManager->>LookupPeerStates: mark_unresponsive(peer)
        else Success
            RemotePeer-->>DhtNetworkManager: NodesFound(nodes, seq)
            DhtNetworkManager->>LookupPeerStates: mark_succeeded(peer)
            DhtNetworkManager->>DhtNetworkManager: merge_gossiped_typed_addresses (seq-aware)
        end
    end
    DhtNetworkManager->>DhtNetworkManager: apply_lookup_report_winners(best_nodes, subject_reports)
    DhtNetworkManager-->>Caller: "Vec<DHTNode> (refreshed)"
```

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `src/dht_network_manager.rs`, line 996-1005 ([link](https://github.com/saorsa-labs/saorsa-core/blob/a7380fea86e964adf039f7a1c2b4529c4da656ac/src/dht_network_manager.rs#L996-L1005)) 

   <a href="#">P2</a> **Lagged broadcast receiver silently loses the failure signal**

   When the broadcast ring buffer overflows a `Lagged` error is returned and the receiver position is bumped forward, dropping the skipped messages. If the specific `peer_id` this waiter is watching was among the dropped notifications, the loop will keep calling `recv().await` for new messages that never carry that peer, so the `external_failure` branch of the `select!` never fires. The early-abort optimisation is silently lost for that probe for the rest of its lifetime. The `request` branch still resolves the outer future, so there is no deadlock — only a missed short-circuit. Given the 1024-entry buffer this is unlikely in practice, but under a lookup storm where many probes fail in rapid succession the lag window can be wide enough to drop the target.

   `````markdown
   Path: src/dht_network_manager.rs
   Line: 996-1005

   Comment:
   **Lagged broadcast receiver silently loses the failure signal**

   When the broadcast ring buffer overflows a `Lagged` error is returned and the receiver position is bumped forward, dropping the skipped messages. If the specific `peer_id` this waiter is watching was among the dropped notifications, the loop will keep calling `recv().await` for new messages that never carry that peer, so the `external_failure` branch of the `select!` never fires. The early-abort optimisation is silently lost for that probe for the rest of its lifetime. The `request` branch still resolves the outer future, so there is no deadlock — only a missed short-circuit. Given the 1024-entry buffer this is unlikely in practice, but under a lookup storm where many probes fail in rapid succession the lag window can be wide enough to drop the target.

   `````
   </details>

</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
Cargo.toml:67
**Mutable git branch as a production dependency**

`saorsa-transport` now resolves from a live branch (`fix/stability-improvements`) rather than a published registry version or a pinned tag. The Cargo.lock records a specific commit hash, so builds are reproducible today, but the branch can be force-pushed or deleted at any time — any workflow that drops the lock file (e.g., a `cargo update`, a fresh worktree, or a CI job that regenerates the lock) will silently pull whatever HEAD the branch points to at that moment, or fail entirely if the branch no longer exists. Prefer a git tag or a published crate version once the upstream fixes land.

### Issue 2 of 2
src/dht_network_manager.rs:996-1005
**Lagged broadcast receiver silently loses the failure signal**

When the broadcast ring buffer overflows a `Lagged` error is returned and the receiver position is bumped forward, dropping the skipped messages. If the specific `peer_id` this waiter is watching was among the dropped notifications, the loop will keep calling `recv().await` for new messages that never carry that peer, so the `external_failure` branch of the `select!` never fires. The early-abort optimisation is silently lost for that probe for the rest of its lifetime. The `request` branch still resolves the outer future, so there is no deadlock — only a missed short-circuit. Given the 1024-entry buffer this is unlikely in practice, but under a lookup storm where many probes fail in rapid succession the lag window can be wide enough to drop the target.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["chore(deps): switch \`saorsa-transport\` t..."](https://github.com/saorsa-labs/saorsa-core/commit/a7380fea86e964adf039f7a1c2b4529c4da656ac) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31460677)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,9 +307,9 @@ checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "cc"
-version = "1.2.61"
+version = "1.2.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d16d90359e986641506914ba71350897565610e87ce0ad9e6f28569db3dd5c6d"
+checksum = "a1dce859f0832a7d088c4f1119888ab94ef4b5d6795d1ce05afb7fe159d79f98"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -1142,9 +1142,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.17.0"
+version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
+checksum = "ed5909b6e89a2db4456e54cd5f673791d7eca6732202bbf2a9cc504fe2f9b84a"
 
 [[package]]
 name = "heapless"
@@ -1469,7 +1469,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d466e9454f08e4a911e14806c24e16fba1b4c121d1ea474396f396069cf949d9"
 dependencies = [
  "equivalent",
- "hashbrown 0.17.0",
+ "hashbrown 0.17.1",
  "serde",
  "serde_core",
 ]
@@ -1587,9 +1587,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.97"
+version = "0.3.98"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1840c94c045fbcf8ba2812c95db44499f7c64910a912551aaaa541decebcacf"
+checksum = "67df7112613f8bfd9150013a0314e196f4800d3201ae742489d999db2f979f08"
 dependencies = [
  "cfg-if",
  "futures-util",
@@ -2628,8 +2628,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.34.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31ebc09fb51e2c325e61ff09892f1256c60ef4f3beb881fd2980befe3e53e9bc"
+source = "git+https://github.com/saorsa-labs/saorsa-transport?branch=fix%2Fstability-improvements#55f423ca5e3312e31ba22475b68a76f4ffd50285"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3118,9 +3117,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.52.2"
+version = "1.52.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "110a78583f19d5cdb2c5ccf321d1290344e71313c6c37d43520d386027d18386"
+checksum = "8fc7f01b389ac15039e4dc9531aa973a135d7a4135281b12d7c1bc79fd57fffe"
 dependencies = [
  "bytes",
  "libc",
@@ -3185,9 +3184,9 @@ dependencies = [
 
 [[package]]
 name = "tower-http"
-version = "0.6.9"
+version = "0.6.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a28f0d049ccfaa566e14e9663d304d8577427b368cb4710a20528690287a738b"
+checksum = "68d6fdd9f81c2819c9a8b0e0cd91660e7746a8e6ea2ba7c6b2b057985f6bcb51"
 dependencies = [
  "bitflags",
  "bytes",
@@ -3434,9 +3433,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.120"
+version = "0.2.121"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df52b6d9b87e0c74c9edfa1eb2d9bf85e5d63515474513aa50fa181b3c4f5db1"
+checksum = "49ace1d07c165b0864824eee619580c4689389afa9dc9ed3a4c75040d82e6790"
 dependencies = [
  "cfg-if",
  "once_cell",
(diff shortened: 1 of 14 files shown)


mickvandijke:
@claude Please review

grumbach (approved):
Approving. Corroborated independently on the 210-node 7-region anselme-testnet 2026-05-12 — DHT formed and stabilised cleanly with the full stability bundle. Chris's home-connection upload run and his 900-node DEV-01 testnet last weekend both showed clear wins (per his 2026-05-11 Slack signoff: *"2-3x increase in download speed and a significant increase in upload speed"*).

This is #2 in the merge chain (transport -> core -> protocol -> node -> client).

PR #117, opened 2026-05-15
fix(dht): drop wildcard / port-zero addresses at store side

grumbach:
## Why

Prod fleet measurement (24 ant-node v0.11.3 hosts, 3h47m window, 2026-05-14 23:15 → 2026-05-15 03:02 UTC) shows **66,653** of these WARN lines:

```
WARN saorsa_core::dht_network_manager: Rejecting unspecified address: /ip6/::/udp/<port>/quic
```

That's the loudest WARN pattern in the entire log corpus and represents 58% of all WARNs across the fleet. Every prod node sees it.

### Root cause

Older saorsa-core builds (pre-self-address filter at `self_address.rs:257`) publish their bind-side wildcard addresses verbatim via `PUBLISH_ADDRESS_SET`, `OBSERVED_ADDRESS`, and `FIND_NODE` gossip. Receivers store them as if they were dialable, then **every** dial attempt against the peer reads the wildcard back out of the routing table and emits the WARN at `dht_network_manager.rs:2537`. Once a wildcard enters our routing table from a stale-version peer, it sticks — there's no expiry that drops "addresses that fail `is_dialable` permanently."

The current self-side publish filter (`self_address.rs:257`) blocks our own node from broadcasting wildcards, but does nothing for wildcards we receive from peers.

## What

Adds a centralised `is_storable_address(&MultiAddr) -> bool` helper that returns `false` for `0.0.0.0`, `::`, and port-zero on any IP-based transport (QUIC, TCP, UDP), and applies it at the four store entry points:

- **`NodeInfo::merge_typed_address`** — returns silently
- **`NodeInfo::merge_typed_address_upgrade_only`** — returns `false` (no change reported)
- **`KBucket::add_node`** — filters incoming `node.addresses`; rejects whole record if all addresses are stripped
- **`KBucket::replace_node_addresses_with_mode`** — filters input vec. **If filtering empties a non-empty input, returns `false`** so the caller (`KademliaRoutingTable::replace_node_addresses_with_mode`) does not advance `last_publish_seqs` and does not wipe the peer's prior good addresses. This preserves the peer's working state AND keeps the publish-sequence window open for a subsequent correct republish at the same seq.

The downstream `is_dialable` check in `DhtNetworkManager` keeps its existing rejection so any wildcard slipping past a future store path is still caught at dial time (defence in depth).

### What's intentionally NOT filtered

- **Loopback** (`127.0.0.1` / `::1`): local devnets legitimately store loopback addresses, and the per-node `allow_loopback` config gates routing-table admission separately.
- **Non-IP transports**: the address type itself encodes whatever reachability semantics are appropriate.

## Tests

13 new tests added covering:

- The helper (5 tests: wildcard v4/v6, port zero, routable accept, loopback accept)
- Both `NodeInfo` merge variants (2 tests)
- Both `KBucket` entry points (3 tests)
- TCP and UDP wildcards (not just QUIC) (2 tests)
- The preservation invariant: filter-empty publish must not wipe good state (1 test)
- Mixed Relay + Direct + wildcard publish (1 test)

All 475 lib tests pass. `cargo fmt --all -- --check` clean. `cargo clippy --lib --all-features -- -D warnings` clean.

## Adversarial review

Ran an adversarial review pass before push. Two issues found and fixed:

1. **TCP / raw-UDP wildcards bypassed the filter** — `dialable_socket_addr()` only matches QUIC. Switched to `socket_addr()` to catch all IP transports.
2. **Empty filtered publish would have wiped good state** — original draft would have applied an empty replace and advanced the publish sequence, locking out future correct republishes at the same seq. Now returns `false` early so the caller skips the seq advance.

## Impact

Expected: silences ~60k WARN/h across the fleet immediately on rollout, with no eviction or behaviour change. Routing table entries for misbehaving peers no longer accumulate dead wildcard addresses, so dial budgets aren't wasted on them. No change to dial logic, no change to admission, no change to any wire format.

<h3>Greptile Summary</h3>

This PR adds a `is_storable_address` guard that drops wildcard (`0.0.0.0` / `::`) and port-0 addresses at four routing-table store entry points, silencing the ~66k WARN/h seen on the prod fleet where older peers published their bind-side addresses verbatim. A preservation guard in `replace_node_addresses_with_mode` returns `false` (without advancing `last_publish_seqs`) when the entire input is filtered to empty, preventing good addresses from being wiped while keeping the sequence window open for a correct republish.

- `is_storable_address` uses `socket_addr()` (not `dialable_socket_addr()`) so TCP and bare-UDP wildcards are also caught, not just QUIC.
- The preservation invariant is well-tested (13 new tests), but the early-return path in `replace_node_addresses_with_mode` skips `last_seen.store_now()` for `AuthenticatedSelfPublish` mode, which is a subtle behaviour change from the stated \"no eviction\" goal for peers that publish exclusively wildcard addresses.

<h3>Confidence Score: 4/5</h3>

Safe to merge for the primary goal of silencing the wildcard-address WARN flood; the two issues are minor doc staleness and an edge-case liveness side-effect.

The core logic is correct and well-tested across 13 scenarios including the critical preservation invariant. The liveness side-effect only matters for peers that exclusively publish wildcard addresses, which are effectively unreachable and can still refresh liveness via FIND_NODE responses.

src/dht/core_engine.rs — specifically the replace_node_addresses_with_mode early-return path and the stale doc comment.

| Filename | Overview |
|----------|----------|
| src/dht/core_engine.rs | Adds `is_storable_address` helper and applies it at four store entry points; includes a preservation guard that returns false without advancing last_publish_seqs when all addresses in a non-empty publish are filtered; 13 new unit tests cover the key paths. Minor: doc comment on replace_node_addresses not updated for the new false case, and the early return for all-wildcard AuthenticatedSelfPublish skips the last_seen refresh. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A["PublishAddressSet received"] --> B["filter_lan_addresses_for_store"]
    B --> C["DhtCoreEngine::replace_node_addresses"]
    C --> D{"typed_addresses empty?"}
    D -- "yes" --> E["return false"]
    D -- "no" --> F["filter_publish_address_set"]
    F --> G["KademliaRoutingTable::replace_node_addresses_with_mode"]
    G --> H{"seq invalid/stale?"}
    H -- "yes" --> I["return false"]
    H -- "no" --> J["KBucket::replace_node_addresses_with_mode"]
    J --> K["filter is_storable_address"]
    K --> L{"input non-empty, result empty?"}
    L -- "yes" --> M["return false: addresses preserved, seq not advanced, last_seen not refreshed"]
    L -- "no" --> N["sort + truncate + cap"]
    N --> O["write node.addresses"]
    O --> P{"AuthenticatedSelfPublish?"}
    P -- "yes" --> Q["last_seen.store_now, advance seq"]
    P -- "no" --> R["advance seq only"]
```

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `src/dht/core_engine.rs`, line 782-783 ([link](https://github.com/saorsa-labs/saorsa-core/blob/6ec50a0701dc04e8526136388d69b3f0393c335b/src/dht/core_engine.rs#L782-L783)) 

   <a href="#">P2</a> The doc comment for `replace_node_addresses` (and the private `replace_node_addresses_with_mode`) now has a stale return-value description. The comment says `false` means "peer is not in this bucket," but after this PR there is a second reason to return `false`: a non-empty input set that is entirely filtered to wildcards. A caller that checks the return value to infer bucket membership will draw the wrong conclusion in the new case.

   

   `````markdown
   Path: src/dht/core_engine.rs
   Line: 782-783

   Comment:
   The doc comment for `replace_node_addresses` (and the private `replace_node_addresses_with_mode`) now has a stale return-value description. The comment says `false` means "peer is not in this bucket," but after this PR there is a second reason to return `false`: a non-empty input set that is entirely filtered to wildcards. A caller that checks the return value to infer bucket membership will draw the wrong conclusion in the new case.

   

   `````
   </details>

</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
src/dht/core_engine.rs:782-783
The doc comment for `replace_node_addresses` (and the private `replace_node_addresses_with_mode`) now has a stale return-value description. The comment says `false` means "peer is not in this bucket," but after this PR there is a second reason to return `false`: a non-empty input set that is entirely filtered to wildcards. A caller that checks the return value to infer bucket membership will draw the wrong conclusion in the new case.

```suggestion
    /// Returns `true` when the peer was found and its addresses replaced,
    /// `false` when the peer is not in this bucket or when every address in
    /// `typed_addresses` was filtered out as non-storable (e.g., all wildcards).
```

### Issue 2 of 2
src/dht/core_engine.rs:838-840
**Liveness not refreshed on all-wildcard authenticated self-publish**

The early `return false` here exits before the `if mode.refreshes_liveness()` block that calls `last_seen.store_now()` and moves the node to the tail (lines 858–863). For `AuthenticatedSelfPublish` mode, this means a peer that sends a `PublishAddressSet` containing only wildcards will not have its `last_seen` timestamp updated even though it demonstrably just communicated with us. The PR description says "no eviction or behaviour change," but this path silently stops refreshing liveness for that edge case. A minimal fix would be to separate the address-update return value from the liveness-update side effect, or to explicitly document that this is intentional.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(dht): drop wildcard / port-zero addr..."](https://github.com/saorsa-labs/saorsa-core/commit/6ec50a0701dc04e8526136388d69b3f0393c335b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32274191)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/src/dht/core_engine.rs
+++ b/src/dht/core_engine.rs
@@ -251,6 +251,33 @@ impl NodeInfo {
             .collect()
     }
 
+    /// Drop every non-storable address (wildcard `0.0.0.0` / `::`, port-zero)
+    /// in place, keeping `addresses` and `address_types` index-aligned.
+    ///
+    /// This is the record-level equivalent of the single-address filter in
+    /// [`Self::merge_typed_address`]. Admission paths that derive diversity
+    /// state from the raw address list (e.g. `DhtCoreEngine::add_node`
+    /// building `candidate_ips`) must call this first so a record carrying
+    /// both a wildcard and a routable address is admitted with only the
+    /// routable address, instead of being rejected wholesale by the
+    /// unspecified-IP guard before the bucket-side store filter can run.
+    pub fn retain_storable_addresses(&mut self) {
+        // Pad address_types so the index alignment holds before we filter
+        // (legacy records may carry fewer tags than addresses).
+        while self.address_types.len() < self.addresses.len() {
+            self.address_types.push(AddressType::Unverified);
+        }
+        // Rebuild both vectors in lockstep, keeping only storable addresses.
+        let addresses = std::mem::take(&mut self.addresses);
+        let address_types = std::mem::take(&mut self.address_types);
+        for (addr, addr_type) in addresses.into_iter().zip(address_types) {
+            if is_storable_address(&addr) {
+                self.addresses.push(addr);
+                self.address_types.push(addr_type);
+            }
+        }
+    }
+
     /// Merge a new address with default type `Direct`.
     /// Prefer `merge_typed_address` when the type is known.
     pub fn merge_address(&mut self, addr: MultiAddr) {
@@ -1736,18 +1763,30 @@ impl DhtCoreEngine {
     /// the same replacement through [`Self::replace_node_addresses_from_gossip`]
     /// without refreshing liveness.
     ///
-    /// - Loopback entries are filtered out unless [`Self::allow_loopback`]
-    ///   is set (devnets/tests), matching the loopback-injection guard in
-    ///   [`KBucket::touch_node_typed`].
-    /// - Empty input address lists are rejected. If filtering strips every
-    ///   supplied address, the empty set is still applied so stale addresses
-    ///   from an older publish cannot survive a newer full replacement.
+    /// - Empty input address lists are rejected outright (returns `false`).
+    /// - Filtering happens in two stages with **different empty-result
+    ///   semantics**, because they guard against different things:
+    ///   - *Loopback* (this method, via [`Self::filter_publish_address_set`],
+    ///     skipped when [`Self::allow_loopback`] is set for devnets/tests):
+    ///     if stripping loopback empties a non-empty input, the **empty set
+    ///     is still applied**. A publisher that legitimately drops to zero
+    ///     reachable addresses must not keep stale ones alive, and
+    ///     `last_publish_seqs` advances normally.
+    ///   - *Wildcard / port-zero* ([`KBucket::replace_node_addresses_with_mode`]
+    ///     via `is_storable_address`): if this empties a non-empty input the
+    ///     replace is **refused** — prior good addresses are preserved and
+    ///     `last_publish_seqs` is NOT advanced, so a corrected republish at
+    ///     the same `seq` can still land. Wildcard-only is treated as a
+    ///     malformed publish, not an intentional "I have no addresses", so
+    ///     callers and tests must not assume all filter-empty publishes
+    ///     share the same outcome.
     /// - `seq` must be non-zero and strictly exceed the last sequence
     ///   observed from `node_id`; zero, older, or duplicate sequences are
     ///   ignored.
     ///
     /// Returns `true` when the peer's addresses were replaced, `false`
-    /// otherwise (peer absent, stale sequence, or empty input list).
+    /// otherwise (peer absent, stale sequence, empty input list, or a
+    /// wildcard-only input whose prior addresses were preserved).
     pub async fn replace_node_addresses(
         &self,
         node_id: &PeerId,
@@ -1827,7 +1866,7 @@ impl DhtCoreEngine {
     /// (DhtNetworkManager) must handle the revalidation flow.
     pub async fn add_node(
         &mut self,
-        node: NodeInfo,
+        mut node: NodeInfo,
         trust_score: &impl Fn(&PeerId) -> f64,
     ) -> Result<AdmissionResult> {
         // Reject self-insertion — a node must never appear in its own routing table.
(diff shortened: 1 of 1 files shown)


mickvandijke (changes requested):
Great find and fix!!

Codex just had two issues that seem worth addressing:

Findings:

- P2: DhtCoreEngine::add_node can still reject mixed wildcard+routable records before the new store filter runs. The PR filters in KBucket::add_node, but DhtCoreEngine::add_node first builds candidate_ips from the original node.addresses, then add_with_diversity rejects if any candidate IP is unspecified. So a NodeInfo with both /ip4/0.0.0.0/... and a valid routable address is rejected before KBucket::add_node can strip the bad address. The new unit test covers KBucket directly, not the public engine admission path. Please sanitize with is_storable_address before computing candidate_ips, or add an engine-level test proving mixed records are admitted with only the routable address stored.

- P3: The replacement docs now describe stale behavior for wildcard-filtered publishes. DhtCoreEngine::replace_node_addresses still says that if filtering strips every supplied address, the empty set is applied. That remains true for loopback filtering, but the new bucket-side wildcard/port-zero filter returns false and preserves prior addresses. Please document that distinction so callers and tests do not assume all filter-empty publishes have the same semantics.

grumbach:
Thanks @mickvandijke, both addressed in e4baa75:

**P2** — `DhtCoreEngine::add_node` now calls `node.retain_storable_addresses()` before `candidate_ips` is built, so a mixed wildcard+routable record is admitted with only the routable address instead of being rejected by the unspecified-IP guard in `add_with_diversity`. A record left address-empty after sanitization is now rejected explicitly rather than slipping into the non-IP-transport bypass. Added two engine-level tests through the public path: `engine_add_node_admits_mixed_wildcard_and_routable` and `engine_add_node_rejects_only_wildcard_record`.

**P3** — Rewrote the `replace_node_addresses` docs to call out the two distinct empty-result semantics: loopback-filter-empty applies the empty set (intentional 'no addresses', seq advances), wildcard/port-zero-filter-empty refuses the replace and preserves prior addresses without advancing `last_publish_seqs`.

Full `dht::core_engine` suite green (106 passed), fmt + clippy clean.

mickvandijke (approved):
LGTM!

PR #123, opened 2026-05-28
fix(p2p): remove wall-clock timestamp validation from WireMessage

dirvine:
## Problem

Clock skew between peers (>5 min) was silently dropping P2P traffic. Consumer devices commonly drift by minutes or hours (wrong timezone, suspended laptops, VMs), and testnets are likely mostly UTC so this was masked.

This is part of the Autonomi 2.0 issue where clients on non-UTC machines could not reliably download/upload.

## Change

- Remove timestamp-based acceptance/rejection from `parse_protocol_message()` in `src/network.rs`.
- Keep timestamp in `WireMessage` and in the signed payload for integrity.
- Remove `MAX_MESSAGE_AGE_SECS` and `MAX_FUTURE_SECS` constants.
- Keep the transport/authenticated peer as the message source; the payload `from` remains protected by signature verification for signed messages.
- Add regression coverage for arbitrary old/future timestamps on both unsigned and signed messages.
- Fix the displaced doc comment so it documents `parse_protocol_message`, not `protocol_error`.

## Security note

This deliberately removes wall-clock replay gating at the generic wire-message parser. That is the point of the fix: message acceptance must not depend on local wall-clock alignment between peers. Signature verification still covers the timestamp, payload, protocol, sender, and user agent for integrity/authenticity. Replay-sensitive semantics should be handled at the relevant protocol/application layer rather than by rejecting all skewed P2P messages globally.

## Verification

- `cargo test test_parse_protocol_message_accepts_arbitrary_timestamps --all-features` — passed.
- `cargo test --all-features --no-fail-fast` — 482 unit tests passed; integration/doc tests passed; known ignored tests remained ignored.
- `cargo clippy --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used` — passed.

## Related

- Autonomi 2.0 client upload/download failures on machines not set to UTC.
- Paired with WithAutonomi/ant-node#120, which removes payment quote wall-clock freshness and replaces it with storage-delta validation.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -1723,16 +1723,6 @@ impl P2PNode {
     }
 }
 
-/// Parse a postcard-encoded protocol message into a `P2PEvent::Message`.
-///
-/// Returns `None` if the bytes cannot be deserialized as a valid `WireMessage`.
-///
-/// The `from` field is a required part of the wire protocol but is **not**
-/// used as the event source. Instead, `source` — the transport-level peer ID
-/// derived from the authenticated QUIC connection — is used so that consumers
-/// can pass it directly to `send_message()`. This eliminates a spoofing
-/// vector where a peer could claim an arbitrary identity via the payload.
-///
 /// Convenience constructor for `P2PError::Network(NetworkError::ProtocolError(...))`.
 fn protocol_error(msg: impl std::fmt::Display) -> P2PError {
     P2PError::Network(NetworkError::ProtocolError(msg.to_string().into()))
@@ -1755,6 +1745,15 @@ pub(crate) struct ParsedMessage {
     pub(crate) user_agent: String,
 }
 
+/// Parse a postcard-encoded protocol message into a `P2PEvent::Message`.
+///
+/// Returns `None` if the bytes cannot be deserialized as a valid `WireMessage`.
+///
+/// The `from` field is a required part of the wire protocol but is **not**
+/// used as the event source. Instead, `source` — the transport-level peer ID
+/// derived from the authenticated QUIC connection — is used so that consumers
+/// can pass it directly to `send_message()`. This eliminates a spoofing
+/// vector where a peer could claim an arbitrary identity via the payload.
 pub(crate) fn parse_protocol_message(bytes: &[u8], source: &str) -> Option<ParsedMessage> {
     let message: WireMessage = postcard::from_bytes(bytes).ok()?;
     let transport_source = source.parse::<SocketAddr>().ok().map(MultiAddr::quic);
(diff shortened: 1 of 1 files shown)


mickvandijke (approved):

dirvine:
Addressed Mick/Greptile follow-up on the saorsa-core PR:

- Added signed-message coverage to `test_parse_protocol_message_accepts_arbitrary_timestamps`; it now checks both unsigned and signed WireMessages with timestamps 10h in the past/future.
- Kept timestamp inside the signed payload for integrity, but confirmed parser acceptance no longer depends on wall-clock age/skew.
- Fixed the displaced doc comment so it documents `parse_protocol_message`, not `protocol_error`.
- Updated the PR body to remove stale ant-node wording and link the paired fix to WithAutonomi/ant-node#120.

Retested locally:
- `cargo test test_parse_protocol_message_accepts_arbitrary_timestamps --all-features` — passed
- `cargo test --all-features --no-fail-fast` — passed: 482 unit tests; integration/doc tests pass; known ignored tests remain ignored
- `cargo clippy --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used` — passed

grumbach:
## Blocker

- `src/network.rs:1739-1800` — after removing the in-parser skew checks, `parse_protocol_message()` no longer consumes `message.timestamp` for any policy decision, but it also does not return it in `ParsedMessage` or the constructed `P2PEvent::Message`. Before this PR, replay/clock policy existed here; after this PR, accepted messages carry no timestamp metadata at all. That makes downstream replay defense impossible to add on top of the parser/broadcast path without another wire-format parse or API change. Suggested fix: expose the signed timestamp on `ParsedMessage` / `P2PEvent::Message` in this PR so handlers can implement freshness/dedup later.

## Concern

- `src/network.rs:3514-3551` — the new regression test mostly proves `parse_protocol_message()` returns `Some`. For signed cases it doesn't assert that `authenticated_node_id` / `P2PEvent::Message.source` are still populated, and it doesn't prove the stated invariant that the timestamp is part of the signed bytes. A stronger test would sign once, mutate only `timestamp`, and assert rejection.

## Positives

- `parse_protocol_message()` still preserves the important identity/routing behavior: unsigned messages do not trust wire `from`, signed messages still go through `verify_message_signature()`, and `transport_source` is still derived independently from the transport address.
- The doc comment is now attached to the parser instead of the removed constants — fixes the misplaced documentation.

## Question

- If replay handling is intentionally moving out of `parse_protocol_message()`, what consumer is supposed to receive the timestamp? Worth exposing it on `ParsedMessage` / `P2PEvent::Message` in this PR to avoid locking the API into accepted-but-impossible-to-reason-about semantics.

PR #129, opened 2026-06-11
feat(dht): relay-IP-level dial-failure inference in DialFailureCache

jacderida:
## Context

Investigation of the PROD-UL-01 merkle upload run (a single 4 MB-chunk uploader on `lon1`) found that slow uploads and quorum-short failures were dominated by **iterative DHT lookups dialling dead address records**. NAT traversal itself was healthy — the problem was the *breadth* of stale records each lookup surfaced and probed; almost all dial failures were **dead records, not slow/marginal paths**.

Grouping dead relay-session addresses by **relay-server IP** revealed a structural pattern: when a relay server goes down it orphans **every session address it ever allocated**, and those records keep circulating in FIND_NODE responses. The existing per-`SocketAddr` `DialFailureCache` can't collapse them — each orphaned session is a distinct `ip:port`, so each must fail its own dial before being suppressed.

## Change

Adds an **IP-level failure tier** alongside the existing per-`SocketAddr` tier in `DialFailureCache` (`src/dht_network_manager.rs`):

- Tracks, per relay-server IP, the set of **distinct** failed relay-session `SocketAddr`s within the TTL window.
- When that set reaches `DIAL_FAILURE_IP_SUPPRESS_THRESHOLD` (4) with no intervening success, **all** remaining sessions at that IP are suppressed for the cache TTL — `is_failed` returns true for any relay `SocketAddr` whose IP is suppressed.
- Scoped to `AddressType::Relay`: only relay failures feed the tier and only relay dials are suppressed, so a genuinely-distinct `Direct` peer behind a shared NAT IP is never suppressed by relay churn there.
- Any successful dial at an IP clears its IP-level suppression immediately, so a recovered relay (or a live peer behind a shared IP) is re-admitted promptly.

Same `DashMap`-backed, lazy-expiry, 30-min TTL design as the existing tier — just a second index. `record_failure` / `is_failed` gain an `AddressType` parameter.

## Test coverage

6 new unit tests: threshold suppression of an unseen session; re-dialling the same session does not inflate the distinct count; success clears suppression; **shared-IP scenario where one peer fails and another succeeds does not suppress the live peer**; non-relay failures never build IP suppression; IP-tier TTL expiry.

## Testnet validation

Validated on a DEV-02 testnet (69% NAT-simulated, fix on both nodes and the `ant` client). A relay-rich public node serving **36 relay sessions** was hard-powered-off and the client's dialling was read from its trace log.

**The win** — the upload spanning the relay's death dialled **4** distinct dead sessions (the threshold), then **skipped 13 more without dialling them**, all 13 confirmed *skipped-but-never-dialled* (true IP-tier collapse): **13 of 17 surfaced dead sessions avoided (~76%)**.

**Safety held** — while the relay was alive the client made **36 successful relay connects with zero suppression** (the success-clear kept the live relay fully usable). Both post-kill uploads completed OK; **no store failures / quorum shortfalls**.

## Known caveats / follow-ups

- **Live-relay false positive (shared-IP mode):** a *healthy* relay that fronts unreachable/stale NAT'd peers can be transiently IP-suppressed, because a relay-session dial failure is charged to the relay *server's* IP even when the real fault is the peer behind the session. Observed at small scale, self-correcting via the success-clear, and caused no store failures — but worth a follow-up: e.g. only counting server-down-looking failures, a higher threshold, or a stickier success-clear.
- **Not a hard cap under concurrency:** α-parallel lookups can launch a burst of dials to a dead IP before the cache populates, so a single process's dialled count can briefly exceed the threshold; suppression then engages for subsequent dials. The benefit is maximised in long-lived processes (nodes, sustained clients) where the cache persists across the full TTL.

### DIFF
--- a/src/dht_network_manager.rs
+++ b/src/dht_network_manager.rs
@@ -845,6 +845,16 @@ impl DialFailureCache {
         Self::default()
     }
 
+    /// Canonicalize a cache key so an IPv4-mapped IPv6 socket
+    /// (`[::ffff:a.b.c.d]:p`) and its bare IPv4 form (`a.b.c.d:p`) resolve to
+    /// the **same** entry — matching how the rest of the dial path
+    /// canonicalizes endpoint identity (see [`canonicalize_ip`]). Without
+    /// this, a failure recorded under one form would neither suppress nor be
+    /// cleared by the other for the same physical endpoint.
+    fn canon_key(addr: SocketAddr) -> SocketAddr {
+        SocketAddr::new(canonicalize_ip(addr.ip()), addr.port())
+    }
+
     /// Returns true if `addr` should be skipped because it failed a dial
     /// within the last [`DIAL_FAILURE_CACHE_TTL`] (per-`SocketAddr` tier),
     /// or — for `AddressType::Relay` dials only — because its relay-server
@@ -857,7 +867,8 @@ impl DialFailureCache {
     /// [`Self::record_failure`] cannot slip a fresh entry in between
     /// the check and the eviction.
     fn is_failed(&self, addr: &SocketAddr, ty: AddressType) -> bool {
-        let per_addr_failed = match self.entries.entry(*addr) {
+        let addr = Self::canon_key(*addr);
+        let per_addr_failed = match self.entries.entry(addr) {
             DashEntry::Occupied(entry) => {
                 if entry.get().elapsed() < DIAL_FAILURE_CACHE_TTL {
                     true
@@ -904,13 +915,29 @@ impl DialFailureCache {
     /// when `ty` is [`AddressType::Relay`], so enough distinct dead
     /// sessions at one IP trip IP-level suppression.
     fn record_failure(&self, addr: SocketAddr, ty: AddressType) {
+        let addr = Self::canon_key(addr);
         let now = Instant::now();
         self.entries.insert(addr, now);
         if ty == AddressType::Relay {
-            self.ip_failures
-                .entry(addr.ip())
-                .or_default()
-                .insert(addr, now);
+            let mut sessions = self.ip_failures.entry(addr.ip()).or_default();
+            let newly_distinct = sessions.insert(addr, now).is_none();
+            let count = sessions.len();
+            drop(sessions);
+            // Log once — when a newly-distinct failed session first brings the
+            // IP to the suppression threshold — so traces can tell a single
+            // skipped address apart from a fully suppressed relay IP. (The
+            // count may include not-yet-expired entries; `ip_is_suppressed`
+            // prunes on read, so this is best-effort observability.)
+            if newly_distinct && count == DIAL_FAILURE_IP_SUPPRESS_THRESHOLD {
+                debug!(
+                    "DialFailureCache: relay IP {} reached the suppression threshold \
+                     ({} distinct failed sessions, no intervening success); suppressing \
+                     all further relay dials to this IP for up to {:?}",
+                    addr.ip(),
+                    DIAL_FAILURE_IP_SUPPRESS_THRESHOLD,
+                    DIAL_FAILURE_CACHE_TTL
+                );
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


grumbach (approved):
Reviewed the IP-tier design and tests. Scoping to relay-kind dials only, the distinct-session threshold, and the success-clear cover the main false-positive paths, and the shared-IP and TTL-expiry tests pin the safety behaviour. One non-blocking suggestion: a debug log when an IP first crosses the suppression threshold, since prod traces currently can't distinguish a single skipped address from a fully suppressed relay IP.

mickvandijke (commented):
I found two issues in this change.

1. **Medium: cache-exhausted address views are promoted to peer-level lookup failures.**

In `find_closest_nodes_network`, when a candidate's current dial plan is fully cached, we call `peer_states.mark_failed(node.peer_id)` and skip it. That is too strong for a cache-derived decision, especially with this PR's new relay-IP tier: a single responder can give us a stale/suppressed relay-only view of a peer, and that marks the whole peer failed for the rest of the lookup before later responders can contribute better addresses.

Concrete example:

1. The dial failure cache has already suppressed relay IP `198.51.100.10` because four distinct relay-session dials at that IP failed.
2. During an iterative lookup, responder `R1` returns subject peer `P` with only:

```text
P -> /ip4/198.51.100.10/udp/45001/quic  AddressType::Relay
```

`lookup_candidate_dial_plan_is_exhausted(&P)` returns true because that relay IP is suppressed, so the code marks `P` failed and continues.

3. Later in the same lookup, responder `R2` returns the same subject peer `P`, but with a usable address:

```text
P -> /ip4/203.0.113.77/udp/9000/quic  AddressType::Direct
```

4. That second report is skipped before its address can be merged, because `peer_states.is_contactable(&node.peer_id)` is now false for `P`.

The result is that one cached/suppressed address view can poison the peer for the rest of the lookup, even though another responder has a usable address for that same peer. I think cache exhaustion should mean "skip this address view/candidate for now", not a final peer failure, or we should merge reports first and evaluate exhaustion against the combined winning address set.

2. **Low: the new IP tier uses raw socket/IP keys while surrounding code normalizes IPv4-mapped IPv6.**

`record_failure`, `is_failed`/`ip_is_suppressed`, and `clear` key the cache directly with `addr` / `addr.ip()`. Other paths normalize IPv4-mapped IPv6 before comparing/dialing, so failures recorded as `[::ffff:a.b.c.d]:port` will not suppress or be cleared by `a.b.c.d:port` for the same endpoint. This is especially relevant because the address parser accepts `/ip6/::ffff:...` forms. Consider normalizing the `SocketAddr` before cache keying, and canonicalizing the IP key, so the cache matches the transport's endpoint identity.

jacderida:
Thanks both — addressed in `13cf6ba`.

### @mickvandijke

**Issue 1 (cache exhaustion → peer-level failure).** Agreed, and I confirmed the mechanism against the code: `is_contactable` is `!states.contains_key`, so `mark_failed` is terminal, and a later responder's report for the same peer is dropped at the `is_contactable` gate before its addresses can merge. This branch's IP tier amplifies it (a whole relay IP can be suppressed — including the live-relay false positive we hit on a testnet, where a *healthy* relay fronting unreachable NAT'd peers got transiently suppressed).

Fixed by removing the terminal `mark_failed` at **all three** exhaustion sites (initial local set, dequeue-to-query, responder report). Exhaustion now skips the candidate *for this round* but leaves the peer contactable, so a later usable (e.g. `Direct`) address can re-admit it. I took the "skip this view, don't fail the peer" option rather than the merge-first restructure — happy to do the fuller merge-then-evaluate-against-the-combined-set as a follow-up if you'd prefer that.

Honest caveat: no new unit test for this one — `find_closest_nodes_network` has no lookup/mock-transport harness, and the change is a conservative, strictly-more-forgiving removal (it only revives peers that actually have a usable address). Flagging it explicitly rather than papering over it.

**Issue 2 (IPv4-mapped IPv6 keys).** Fixed — `DialFailureCache` now canonicalizes its keys via `canonicalize_ip` (the same helper the surrounding dial path uses) in `record_failure` / `is_failed` / `ip_is_suppressed` / `clear`, so `[::ffff:a.b.c.d]:p` and `a.b.c.d:p` resolve to one entry across both tiers. Added unit tests for the per-address and IP tiers.

### @grumbach

Added the debug log: `record_failure` emits a one-shot `debug!` when a relay IP first crosses the suppression threshold (guarded on the newly-distinct insert that reaches the threshold, so it fires once rather than on every re-dial) — traces can now tell a single skipped address from a fully suppressed relay IP.

---

`cargo test --lib`: 467 passed (incl. the 2 new tests); clippy strict clean.

PR #136, opened 2026-06-17
feat(reachability): gate relay publication with canary quorum

mickvandijke:
## Linear issue
https://linear.app/autonominetwork/issue/V2-812/gate-proactive-relay-publication-with-independent-canary-verification Tracking issue: https://github.com/WithAutonomi/saorsa-core/issues/138. Transport PR: https://github.com/WithAutonomi/saorsa-transport/pull/131.

## Risk tier
- [x] T3 — protocol / storage format / payments / routing. T2 evidence + adversarial testing.

## Compatibility
- Wire: adds the first unreleased `relay-canary-v1` request/response protocol on `/rr/relay-canary-v1`. Requests contain the authenticated target peer ID, public relay socket, and hourly witness-eligibility epoch; they contain no allocation receipt. During mixed-version rollout, a delivered request with no protocol response before the deadline counts as assumed-positive legacy evidence. Witness connection/send failures and explicit unavailable/rate-limit responses remain ineligible; explicit probe failures count as failures.
- Storage: none; DHT record encoding is unchanged.
- API: saorsa-core consumes saorsa-transport's opaque `PreparedRelay` lifecycle. The unreleased relay-allocation receipt surface has been removed from both PRs.

## Semver impact
- [ ] breaking
- [x] feature
- [ ] fix

## Summary

Gate proactive MASQUE relay publication behind independent canary evidence and continuously revalidate established relays. A relay allocation is provisional until admission quorum succeeds; a completed maintenance majority rejection withdraws it immediately. Canary witnesses are a deliberately bounded public dial service and no longer trust requester-mintable receipts.

## Changes

- Add randomized, non-close relay witnesses using the authenticated `relay-canary-v1` request/response protocol.
- Permit a requester to test only its own signed transport identity and accept only public destination sockets.
- Deterministically assign roughly one quarter of witnesses to each target per hourly epoch; witnesses accept the current or immediately previous epoch.
- Bound witness work to 4 requests per target peer/hour, 20 per source IPv4 or IPv6 `/64`/hour, 4 per destination socket/hour, 20 per destination IP/hour, 60 total per witness/hour, and 4 concurrent probes.
- Return only coarse success, failure, or unavailable/rate-limited wire results while retaining detailed failure reasons in local debug logs.
- Require all three admission witnesses to return positive evidence before a provisional relay enters the DHT self-record. A delivered request that times out waiting for a response temporarily supplies assumed-positive compatibility evidence for a legacy witness.
- Require a maintenance majority for established relays; retain an inconclusive result until the ordinary two-hour interval, but immediately withdraw a majority-rejected relay.
- Spread deterministic initial maintenance jitter across the full two-hour interval to avoid fleet-wide canary dial pressure.
- Bind prepare, publish, and abort to the exact opaque transport allocation, preventing stale same-address operations from affecting a replacement relay.
- Keep DHT address publication synchronized with current K-closest peers and retry only pending/new targets.
- Ignore transport relay hints that could bypass the sequenced, canary-gated self-record path.
- Depend on `https://github.com/WithAutonomi/saorsa-transport.git`, branch `fix/pr136-provisional-relay`; `Cargo.lock` pins transport commit `4b1ed67973763e4636694a9d9c0601ff41b9e933` from PR #131.

## Test evidence

- Exact current pair: saorsa-core `ab9da8f` + saorsa-transport `4b1ed679`.
- saorsa-core `cargo test --all-features --quiet`: 560 passed (504 library + 56 integration), 12 ignored, 0 failed; doctests completed.
- saorsa-core `cargo clippy --all-targets --all-features -- -D warnings`: clean against locked transport `4b1ed679`.
- saorsa-transport `cargo test --lib --quiet`: 1,494 passed, 3 ignored, 0 failed.
- saorsa-transport `cargo fmt --all -- --check`: clean.
- saorsa-transport `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- The transport all-feature run passed its 1,580-test primary suite and all receipt/removal-related coverage. Two unrelated NAT coordination race tests failed; both failures reproduce unchanged on an untouched `e24989f2` worktree (6 passed, the same 2 failed).
- downstream ant-node `cargo check --workspace --quiet`: clean with transport lock updated to `4b1ed679` and saorsa-core patched to the current checkout.
- Focused canary suite: 20 passed, covering self-only authentication, current/previous epoch acceptance, stale/ineligible witness rejection, public-address validation, deterministic selection, IPv6 `/64` source bucketing, rate limiting, and quorum semantics.
- The prior receipt-bearing pair (`38e38fa` + `e24989f2`) completed a matched 90-minute comparison with 100 nodes per fleet, 30% NAT, and two continuous 50 MiB uploaders per fleet: main completed 354 uploads and the PR pair 370, with zero failures; PR throughput was +4.5%, upload p50 -6.7%, upload p95 -7.7%, and average node CPU -7.4%. That fleet evidence predates the receipt-removal security hardening and is retained here as transfer-path evidence, not claimed as a run of the exact current commits.
- All 46 comparison droplets were destroyed after collection and both testnet tags were verified empty.

## New dependency
None. The existing saorsa-transport dependency temporarily follows PR #131's Git branch until that PR is released.

## ADR
https://github.com/WithAutonomi/saorsa-core/blob/feat/relay-canary-gate-rc-2026.6.2/docs/adr/ADR-016-canary-gated-proactive-relays.md

## Mitigation / rollback
Revert PR #136 and restore the released saorsa-transport dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#478ee3d36d8016156a8f33bd45844497fd3e4c9a"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#81b809d0c9d6cb5d6eb25c7d024843585e01b68b"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#e24989f21a0bc90cd42bf45ffbdb0bc1700048eb"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#4b1ed67973763e4636694a9d9c0601ff41b9e933"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#4a2ceb96c424f12d76422b06674a45c2472cda71"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#478ee3d36d8016156a8f33bd45844497fd3e4c9a"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,8 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0dbc31faa4bd1f51807dd3815950212005613e512f6b400715067b78c2fb9603"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#4a2ceb96c424f12d76422b06674a45c2472cda71"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#d960473f165107f15e9a7c53dca015bf2033ce17"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#e24989f21a0bc90cd42bf45ffbdb0bc1700048eb"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ dependencies = [
 [[package]]
 name = "saorsa-transport"
 version = "0.35.2"
-source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#81b809d0c9d6cb5d6eb25c7d024843585e01b68b"
+source = "git+https://github.com/WithAutonomi/saorsa-transport.git?branch=fix%2Fpr136-provisional-relay#d960473f165107f15e9a7c53dca015bf2033ce17"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ once_cell = "1.21"
 dashmap = "6"
 
 # Networking
-saorsa-transport = "0.35.2"
+saorsa-transport = { git = "https://github.com/WithAutonomi/saorsa-transport.git", branch = "fix/pr136-provisional-relay" }
 
 # Core-specific dependencies
 dirs = "6.0"
--- a/docs/ROUTING_TABLE_DESIGN.md
+++ b/docs/ROUTING_TABLE_DESIGN.md
@@ -396,10 +396,10 @@ The routing table MUST emit events on membership changes to allow consumers to r
 |---|---|
 | `PeerAdded(PeerId)` | New peer inserted into routing table |
 | `PeerRemoved(PeerId)` | Peer evicted, blocked, or departed |
-| `KClosestPeersChanged { old, new }` | Composition of the `K_BUCKET_SIZE`-closest peers to self changed |
+| `KClosestPeersChanged { old, new, added, removed }` | Composition of the `K_BUCKET_SIZE`-closest peers to self changed |
 | `BootstrapComplete { num_peers }` | Bootstrap process finished (routing table stabilized or timeout reached) |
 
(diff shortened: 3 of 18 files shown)


grumbach:
Agree with the reviews above, both on the canary teardown and on the witness quorum. Nothing to add there.

A few smaller things I did not see raised, none of them blocking:

- `saorsa_transport_adapter.rs`: `DIRECT_CONNECT_TIMEOUT` goes from 1250ms to 3s. Not mentioned in the description, and the comment right above it still says to keep it short so lookups move on quickly. Every unreachable candidate in every lookup now costs 3s.
- `clear_dial_failures_for_published`: skipping addresses already in the previous set does stop the redial loop, but a relay that fails once and then passes maintenance canary at the same address stays suppressed for the full cache TTL. That is the common case once the canary is keeping the address proven.
- Maintenance rejection withdraws after one round. The doc comment in `canary.rs` says established relays confirm the result in a second round before withdrawal, but only `Inconclusive` retries.
- The startup tombstone publishes to `find_closest_nodes_local`, which is near empty at driver start, so it does not reach the replicas actually holding the old record. And when there are no dialable self addresses it publishes an empty address set at a fresh seq, which tells everyone that receives it that the node has no dialable address at all.
- `KClosestPeersChanged` and broadcast lag both clear `last_published_typed_set`, so each one triggers a full K-wide republish. Before this only relayer eviction did that, and K-closest changes are frequent under churn.
- Relay ADD_ADDRESS suppression is gone on the send side in #131 (the `relay_setup_attempted` gates), while the receive side here became a no-op. The DHT_BRIDGE task now drains updates, logs, and does nothing. Worth dropping one side or the other.

dirvine (changes requested):
Reviewed exact head `38e38fa4a2e0ca920c5d621436e4f6749a4cfafb` with a six-seat panel and independent local verification.

## Blocking finding: expired receipts turn maintenance checks into assumed success

A relay allocation receipt expires after 24 hours (`saorsa-transport/src/relay_allocation.rs`, `RECEIPT_LIFETIME_SECS`). The holding driver retains that original receipt and clones it for each two-hour maintenance round (`src/reachability/driver.rs:696-703`).

Once the receipt expires, `verify_relay_canary_receipt` fails at the witness. The current handler logs the rejection and returns without a response (`src/dht_network_manager.rs:4665-4672`). The requester then maps the response timeout to `NoProtocolResponse`, which becomes `AssumedSuccess` (`src/reachability/canary.rs:786-790`). Assumed successes count towards the maintenance quorum.

The result is that a relay held beyond 24 hours can keep passing periodic maintenance with no witness performing the cold dial. Local tunnel-health polling still detects tunnel death, but the external reachability guarantee added by this PR becomes ineffective for long-lived relays.

Please either:

- send an explicit typed non-success response for invalid/expired receipts so it cannot be classified as legacy silence; or
- refresh/reissue the receipt, or deliberately rotate the allocation, before expiry.

Add a regression covering expiry → witness response/classification → maintenance verdict at the production boundary.

## Non-blocking review notes

- The live PR body is now aligned with canary-v2, assumed-success compatibility semantics, and transport head `e24989f2`.
- The git dependency remains unpackageable, but Mick has explicitly assigned the release/version repin to the release manager; I did not treat that as a code-approval blocker.
- The failed Security Audit check is an advisory-database HTTP 503, not an advisory finding. A fresh local `cargo audit --json` found no vulnerability-class advisory.
- Exact-head canary tests pass: 21/21.

Consensus: five substantive seats classify the expiry path as blocking; one seat dissented because it assumed expiry produced a typed rejection and forced reacquisition. Direct inspection refutes that assumption: the handler currently returns silently at `dht_network_manager.rs:4671`.

mickvandijke:
Addressed the blocker from review `4855093059` in core `ab9da8f` with transport `4b1ed679`.

The expired-receipt path no longer exists: allocation receipts and their MASQUE wire/API plumbing were removed entirely. They were not a trustworthy authorization boundary because a requester could generate an ephemeral relayer identity and self-sign a receipt for an arbitrary public address.

`relay-canary-v1` now treats witness probing as an explicitly bounded service:

- the signed transport source must equal the requested target identity;
- a deterministic hourly assignment makes roughly one quarter of witnesses eligible for a target;
- limits are 4/target peer/hour, 20/source IPv4 or IPv6 `/64`/hour, 4/destination socket/hour, 20/destination IP/hour, 60 total/witness/hour, and 4 concurrent probes;
- the source network comes from transport provenance, not request data;
- validation/rate/concurrency rejection returns a typed unavailable result, so it cannot fall through to legacy-timeout assumed success;
- probe responses expose only success, failure, or unavailable/rate-limited; detailed dial/identity errors stay local.

ADR-016 and both PR descriptions now document that behavior and use `relay-canary-v1` (the protocol has not shipped, so there is no v2/v3 lineage).

Validation on the exact pair:

- core: 560 tests passed, 12 ignored, 0 failed; strict all-target/all-feature clippy clean;
- focused canary tests: 20 passed, including epoch boundaries, ineligible witnesses, source-IP bucketing, and quorum classification;
- transport library: 1,494 passed, 3 ignored, 0 failed; strict clippy clean;
- downstream ant-node workspace check: clean with transport locked to `4b1ed679`.

The two unrelated NAT coordination race tests seen in transport's broader all-feature run reproduce unchanged on an untouched `e24989f2` worktree.

dirvine (approved):
Six-seat focused re-review completed on exact core head `ab9da8f90d7b27b5ccf0fdde6c4ba1c3725352fb`, paired with saorsa-transport `4b1ed67973763e4636694a9d9c0601ff41b9e933`.

The prior receipt-expiry blocker is resolved by removing allocation receipts and their MASQUE/API plumbing entirely. Upgraded validation, provenance, rate-limit and concurrency rejection paths now return a typed unavailable result, classified as ineligible rather than positive evidence. The authenticated-self check, hourly witness assignment, source/destination/global budgets and isolated probe ownership were rechecked; no material replacement blocker was found.

Verified:
- focused canary tests: 20 passed;
- exact-head core/transport checks, full library suites and strict clippy clean across the review panel;
- GitHub CI complete with no pending or failed checks;
- PR body and ADR-016 aligned with the receipt-free `relay-canary-v1` design.

Accepted caveats: timeout-to-assumed-success remains the explicit ADR-016 mixed-version rollout trade-off; dependency versioning/packageability remains the release manager's task. Non-blocking follow-ups are to retire the compatibility shim after rollout, update the stale ADR index summary, and add an unknown-response-flag regression test.

Approved; not merged.

PR #145, opened 2026-07-10
feat: add cumulative wire traffic accounting

jacderida:
## Summary

Part of **V2-623** (log-based traffic accounting). DHT/wire messaging overhead
— in particular the ~5.26KB of ML-DSA-65 credentials carried in every
`WireMessage` — is invisible in the logs today. This adds cumulative, monotonic
byte/message counters as an INFO summary line so the telegraf→Elasticsearch
pipeline can measure it. It is the direct before/after instrument for **V2-616**
(stop embedding the ML-DSA-65 public key in every `WireMessage`).
**Purely additive: no wire changes, no behaviour changes.**

- `TrafficCounters` table added to the shared `TransportHandle` (relaxed
  atomics, same style as the sharded dispatcher's `drop_counter`).
- Wire tx + envelope overhead (`wire − payload`) counted at the
  `send_on_channel` choke point, including the identity announce that bypasses
  it; wire rx + overhead counted at the shard-consumer decode site (via a new
  `ParsedMessage.payload_len`).
- Per-message-type counts broken out where the type is known: FIND_NODE,
  `NodesFound`, `PUBLISH_ADDRESS_SET`, ping.
- New `wire traffic summary (cumulative)` line (target `saorsa_core::traffic`)
  emitted from a periodic `DhtNetworkManager` task.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo clippy --all-targets -- -D warnings -D clippy::unwrap_used -D clippy::expect_used`
- [x] `cargo fmt -- --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -552,9 +552,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
--- a/src/dht_network_manager.rs
+++ b/src/dht_network_manager.rs
@@ -38,6 +38,7 @@ use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::net::{IpAddr, SocketAddr};
+use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant, SystemTime};
 use tokio::sync::{RwLock, Semaphore, broadcast, oneshot};
@@ -150,6 +151,12 @@ const MAX_CONCURRENT_REVALIDATIONS: usize = 8;
 
 /// Maximum concurrent pings within a single stale revalidation pass.
 const MAX_CONCURRENT_REVALIDATION_PINGS: usize = 4;
+
+/// Cadence of the `wire traffic summary (cumulative)` INFO line (V2-623).
+/// A `const` so testnets can drop it to 60s; 300s is the production default
+/// that keeps log volume negligible. Emitted at INFO (the shipping pipeline
+/// only carries INFO+).
+const TRAFFIC_SUMMARY_INTERVAL: Duration = Duration::from_secs(300);
 
 /// Minimum self-lookup interval (randomized between min and max).
 const SELF_LOOKUP_INTERVAL_MIN: Duration = Duration::from_secs(300); // 5 minutes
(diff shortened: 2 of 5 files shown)


dirvine (commented):
## Hermes review — telemetry semantics and red gates

The main wire counters are soundly placed: normal TX increments only after `send_on_channel` succeeds, identity announce increments in its success branch, and RX increments after successful decode. Builds and unit tests pass on Linux, macOS and Windows; mutation testing also passes.

One semantic issue needs correction or explicit naming before these fields are used as sent-traffic attribution:

- `find_node_tx_count`, `ping_tx_count` and `publish_addr_tx_count` increment before timestamping, serialisation, channel establishment and send; `nodes_found_tx_*` increments when the response is serialised, before it is transmitted. They therefore count attempts/encoded responses, while `wire_tx_*` counts successful sends. Please move them to a confirmed-send path or rename/document them as attempt counters.

RX totals also exclude malformed/failed-decode traffic; that is reasonable, but the ES reconciliation query should treat them as decoded protocol bytes rather than host ingress.

Current required gates remain red:
- Lint: three dead constants in unchanged `src/validation.rs` (baseline/toolchain drift).
- Security audit: inherited `crossbeam-epoch 0.9.18`, `RUSTSEC-2026-0204` (fixed in `>=0.9.20`).

Local verification: `cargo check --all-targets` passed at head `160f12e906c2191f5e4cad57f28ab7880de0a84e`.

Recommendation: clarify/fix per-type counter semantics, resolve or formally disposition the inherited red gates, then smoke-test one DHT exchange through the production logging pipeline.

jacderida:
Thanks for the review — addressed in `a203189`.

**Per-type counter semantics.** `find_node_tx_count`, `ping_tx_count` and `publish_addr_tx_count` now increment **only after `send_message` succeeds** (matched on `message.payload` in the `Ok(_)` arm), so they share the "confirmed sent" semantics of `wire_tx_*`. Previously they incremented at the funnel entry before serialisation/send.

`nodes_found_tx_*` genuinely counts responses this node *encodes* — the transmit happens later in the caller — so rather than misreport it as confirmed-sent I've documented it explicitly as an encoded/produced-response counter (the wire bytes are still captured by `wire_tx_*` at send time). Happy to move it to the response send site instead if you'd prefer that over the doc clarification.

**RX semantics.** Documented `wire_rx_*` / `overhead_rx_bytes` as decoded protocol bytes (malformed/signature-rejected frames excluded), explicitly not raw host ingress.

**Copilot's `matches!` note.** `matches!(result, ...)` with a `..` pattern doesn't actually move `result` (it compiled and tested fine), but I've switched it to `matches!(&result, ...)` to make the borrow unambiguous.

**Red gates:** the three dead constants in `src/validation.rs` and the inherited `crossbeam-epoch` RUSTSEC-2026-0204 are both pre-existing / inherited (not touched by this PR). I've left them out of scope — glad to take them in a separate change if you want them cleared here.

dirvine (approved):
## Follow-up approval

The requested telemetry-semantic changes are addressed at head `a2031898be9313d735ca597f02f5319d03d3bcd9`:

- FIND_NODE, ping and publish-address counters now increment after successful send.
- `NodesFound` is explicitly documented as produced/encoded rather than confirmed sent.
- RX fields are documented as successfully decoded protocol bytes rather than raw host ingress.

No code-level blocker remains in this logging change. Builds and unit tests pass on Linux, macOS and Windows; mutation testing passes.

Two repository gates remain red outside this diff: baseline dead-code lint and inherited `crossbeam-epoch 0.9.18` / `RUSTSEC-2026-0204`. Approval is for the PR code; those gates still need maintainer disposition before release.

A logging-pipeline smoke test is optional canary validation, not an approval requirement.

PR #150, opened 2026-08-07
chore: relicense from AGPL-3.0 to dual MIT OR Apache-2.0

jacderida:
## Linear issue
V2-849 (rollout) / V2-850 (the licensing decision)

## Risk tier
- [x] T0 — docs / tooling / CI / pure UX-output. Repo CI only.
- [ ] T3 — protocol / storage format / payments / routing. T2 evidence + adversarial testing.

## Compatibility
- Wire: none
- Storage: none
- API: none

## Semver impact
- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence
`cargo check` passes and `cargo fmt --all --check` is clean. No source logic was changed — only licence-grant comment headers and package metadata.

## New dependency
none

## ADR
n/a

## Mitigation / rollback
Revert the PR to restore the previous licence terms in the repository. Note that reverting does not retract a grant already made: anyone who took a copy while the permissive licence was published keeps those rights for that copy. This is inherent to relicensing, not to this change.

## What changed
Implements the licensing decision on V2-850 — `MIT OR Apache-2.0` across all WithAutonomi repositories, moving off copyleft so anyone can build on the code, including in proprietary work.

`LICENSE-MIT` and `LICENSE-APACHE` added at the repo root, copyright **The Autonomi Foundation S/A**.

Relicensed from **AGPL-3.0 plus a commercial licence**. Headers read `AGPL-3.0-or-later OR Commercial` with commercial enquiries directed to david@saorsalabs.com, and the README described the same dual model. Under MIT/Apache-2.0 there is nothing a commercial licence would grant that the permissive licence does not already give away, so that arrangement goes.

It also shipped **no licence file at all** — headers pointed at a `LICENSE-AGPL-3.0` that did not exist in the repo.

35 source files across three header shapes: the AGPL-or-Commercial block (33 files), a longer variant carrying GPL warranty paragraphs (`src/error.rs`, `src/validation.rs`), and a separate AGPL reference block with its own SPDX line (`src/identity/peer_id.rs`, `src/identity/node_identity.rs`). Orphaned "GNU Affero General Public License" warranty paragraphs were removed too.

Also updated: `README.md`, the SPDX identifier in `docs/SECURITY_MODEL.md`, and `AGENTS.md` / `CLAUDE.md` — both instructed contributors to put AGPL headers on new files, so leaving them would have reintroduced AGPL headers on everything new.

---

> The reason for using MIT/Apache-2.0 and moving away from GPL 3.0 is purely allowing anyone to build anything from our code, even proprietary etc. if we are GPL we insist their projects *must* be GPL as well.
>
> Bottom line, builders want it.

Per-file copyright attributions were left as they are throughout — they record who held copyright when the code was written. Only the licence grant text changed.


jacderida:
Merging this without waiting for an approving review.

The decision to move to dual `MIT OR Apache-2.0` has already been confirmed and is recorded on [V2-850](https://linear.app/autonominetwork/issue/V2-850), along with the reasoning — we want people to be able to build anything on our code, including proprietary work, which copyleft prevents. The copyright holder (`The Autonomi Foundation S/A`) was confirmed at the same time.

This PR is one of 17 applying that decision across the org. There is no code change to review: licence files, package metadata, and licence-grant comment headers only. The substantive question here was the licensing choice itself, and that has been settled off the PR rather than in it.

Flagging it explicitly so the merge without a review is on the record rather than looking like an oversight.

